PR #1, opened 2016-03-07
fix(*): fix all import paths

arschles:
To point to `github.com/deis/workflow-cli/...`. Also modifies the `Makefile` to run unit tests and style check commands to run inside a Docker container.

Ref https://github.com/deis/deis/issues/4952

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ _testmain.go
 *.exe
 *.test
 *.prof
+vendor
+deis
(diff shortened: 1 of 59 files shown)


bacongobbler:
Can't add labels because we're not part of the owners group for this repo. Gotta change that in the settings :cry:

jackfrancis:
The `s/workflow/workflow-cli/` instances all look sane. Would be good to have someone else look carefully at the Makefile, but from my view it looks fine.

arschles:
@helgi - I've added core & contributing maintainers to this repo

PR #2, opened 2016-03-07
fix(.travis.yml): modify travis.yml file to be client-specific

arschles:


.travis.yml
```diff
@@ -1,48 +1,25 @@
...
-go:
-  - 1.5.1
-env:
-  - GO15VENDOREXPERIMENT=1
+language: generic
 branches:
   only:
     - master
+env:
+  DEV_REGISTRY=quay.io
```
mboersma: Does this need to be set for CI? The [`deploy.sh` script](https://github.com/deis/workflow-cli/blob/master/_scripts/deploy.sh#L10-L12) sets its own values for `DEV_REGISTRY` as needed.

mboersma: Oh, I see--not used any more. Never mind.

.travis.yml
```diff
@@ -1,48 +1,25 @@
...
-  - make test
-  - make -C client/ bootstrap build test dist prep-bintray-json
+  - make build test dist prep-bintray-json
 deploy:
-  - provider: script
+  - provider: bintray
     # ensure client/doc builds aren't removed
     # see https://docs.travis-ci.com/user/deployment/#Uploading-Files
     skip_cleanup: true
-    script: _scripts/deploy.sh
```
mboersma: We should also remove `_scripts/deploy.sh` if it isn't used now.

arschles: @mboersma good point - done in https://github.com/deis/workflow-cli/commit/cb8ee509338301f0b94a955eb0c4402393c1422c

.travis.yml
arschles: @mboersma good catch - removed in https://github.com/arschles/deis-workflow-cli/commit/3c52ee0011121d07831cc007faaf170b86bec1a5

PR #32, opened 2016-03-31
ref(util): move, export and use SSH public key parsing logic

arschles:
This change will help with https://github.com/deis/workflow-e2e/issues/132


Joshua-Anderson:
Would it maybe make more sense to move this to `pkg/ssh`? IMO it's better to have smaller packages with specific roles than one bigger package with many roles.

mboersma:
> Would it maybe make more sense to move this to `pkg/ssh`?

That's a good point. @arschles has some related refactorings in progress or in mind, so it's his call, but this looks like an improvement on its own so I'm ok with merging it as-is.

arschles:
@mboersma @Joshua-Anderson sorry for the delay! @Joshua-Anderson you have a good point about avoiding the generic `util` package (which can turn into a massive black hole of unrelated code).

I couldn't think of a better name than `util` so I appreciate the suggestion for `pkg/ssh`. I'll make the change shortly.

arschles:
@Joshua-Anderson moved contents of `util` to `pkg/ssh` in https://github.com/deis/workflow-cli/pull/32/commits/34ce414420c1eb9a49675b3c2306a7f2f0ebf4b2

arschles:
this is ready for review, I'd like to get it merged in the beta4 cycle

PR #34, opened 2016-04-06
ref(http.go): improve client error message

mboersma:
This is shamelessly ported directly from @bacongobbler's deis/deis#5013.

This PR improves the resultant client error message they get when the returned response is incorrect. This is especially important for determining the error message

Old message when using a v2 client with a v1 cluster:

```
http://deis.10.245.1.3.xip.io does not appear to be a valid Deis controller. Make sure that the
Controller URI is correct and the server is running.
```

New message:

```
http://deis.10.245.1.3.xip.io/v2/ does not appear to be a valid Deis controller. Make sure that the
Controller URI is correct and the server is running.
```

Notice the appended /v2/ path, which users can assume this is a v2 client.

### DIFF
--- a/controller/client/http.go
+++ b/controller/client/http.go
@@ -156,9 +156,8 @@ func checkForErrors(res *http.Response, body string) error {
 // CheckConnection checks that the user is connected to a network and the URL points to a valid controller.
 func CheckConnection(client *http.Client, controllerURL url.URL) error {
 	errorMessage := `%s does not appear to be a valid Deis controller.
-Make sure that the Controller URI is correct and the server is running.`
-
-	baseURL := controllerURL.String()
+Make sure that the Controller URI is correct, the server is running and
+your client version is correct.`
 
 	controllerURL.Path = "/v2/"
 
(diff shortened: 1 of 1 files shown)


slack:
Good call! Do we also want to add "... Controller URI is correct, the server is running and your client version is correct?"

LGTM either way!

bacongobbler:
Good thought. CI doesn't seem to be happy today so it wouldn't hurt to add more doc checks.

bacongobbler:
oh I thought this was for v1. disregard

bacongobbler:
I'm LGTM'ing this anyways as the code looks good to merge.

carmstrong:
> Do we also want to add "... Controller URI is correct, the server is running and your client version is correct?"

:+1: That was my suggestion in https://github.com/deis/deis/pull/5013#issuecomment-206138783 as well

mboersma:
I added the verbiage @slack suggested so this is in sync with deis/deis#5013.

bacongobbler:
:shipit:

Issue #42, opened 2016-04-15
'deis logs' still references the 'workflow pod'


arschles:
The language should say 'controller pod' instead. Example output (see the 2nd item in the list):

```console
ENG000656:example-go aaronschlesinger$ deis logs
Error: There are currently no log messages. Please check the following things:
1) Logger and fluentd pods are running.
2) If you just installed the logger components via the chart, please make sure you restarted the workflow pod.
3) The application is writing logs to the logger component.
You can verify that logs are appearing in the logger component by issuing the following command:
curl http://<log service ip>:80/logs/gotest on a kubernetes host.
```

bacongobbler:
I'm pretty sure this is a server-side error that's returned, but yes we should fix this up. Should be a one-liner.

bacongobbler:
I lied, it's client-side: https://github.com/deis/workflow-cli/blob/6c176d28c6dc5b8da290e119ecadab744a5645a8/controller/models/apps/apps.go#L100-L109

arschles:
Punting to beta4

arschles:
Moving back to beta3 and closing. #43 fixed this

PR #43, opened 2016-04-15
chore(logger): Update error message when we dont find any log messages

jchauncey:


arschles:
Fixes #42 (incidentally)

controller/models/apps/apps.go
```diff
@@ -101,11 +101,9 @@ func Logs(c *client.Client, appID string, lines int) (string, error) {
...
 			`There are currently no log messages. Please check the following things:
 1) Logger and fluentd pods are running.
-2) If you just installed the logger components via the chart, please make sure you restarted the workflow pod.
-3) The application is writing logs to the logger component.
-You can verify that logs are appearing in the logger component by issuing the following command:
-curl http://<log service ip>:80/logs/%s on a kubernetes host.
-To get the service ip you can do the following: kubectl get svc deis-logger --namespace=deis`, appID)
+2) The application is writing logs to the logger component.
+3) Check that an entry in the ring buffer was created by checking the logger pod logs: kubectl logs <logger pod> --namespace=deis
+4) Making sure that the container logs were mounted properly into the fluentd pod: kubectl exec <fluentd pod> --namespace=deis ls /var/log/containers`, appID)
```
arschles: is there a simple way to indicate what to look for in the above two sets of logs?

jchauncey:
Heres what it looks like - 
```
╰─± deis logs
Error: There are currently no log messages. Please check the following things:
1) Logger and fluentd pods are running.
2) The application is writing logs to the logger component by checking that an entry in the ring buffer was created: kubectl logs <logger pod> --namespace=deis
3) Making sure that the container logs were mounted properly into the fluentd pod: kubectl exec <fluentd pod> --namespace=deis ls /var/log/containers
```

PR #45, opened 2016-04-21
feat(registry): add support for setting private registry information per application

helgi:
ref https://github.com/deis/controller/pull/653

### DIFF
--- a/cmd/registry.go
+++ b/cmd/registry.go
@@ -1,0 +1,128 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/deis/pkg/prettyprint"
+
+	"github.com/deis/workflow-cli/controller/api"
+	"github.com/deis/workflow-cli/controller/models/config"
+)
+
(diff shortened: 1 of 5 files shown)


rimusz:
👍

bacongobbler:
Looks like this is a direct copy of `config`. Why not just make it into a [shortcut](https://github.com/deis/workflow-cli/blob/6c176d28c6dc5b8da290e119ecadab744a5645a8/deis.go#L199) instead?

cmd/registry.go
```diff
@@ -0,0 +1,128 @@
...
+		return err
+	}
+
+	config, err := config.List(c, appID)
+
+	fmt.Printf("=== %s Registry\n", appID)
+
+	registryMap := make(map[string]string)
+
+	for key, value := range config.Registry {
```
bacongobbler: Ah I see here why. Makes sense.

helgi: Same reason as tags, I wanted to do a short cut tho - maybe we can unify them all later?

PR #47, opened 2016-04-23
feat(config): ignore comments in config:push

### LINKED ISSUE #44: Allow commenting out .env vars
When doing a `config:push`, the regex doesn't match and fails pushing variables, while it would be nice to temporarily turn certain `.env` vars off.

Joshua-Anderson:
Fixes #44

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -239,8 +239,16 @@ func ConfigPush(appID, fileName string) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
thanks @Joshua-Anderson! Sorry about the delay; we were just under a code freeze for the past few days before releasing v2.0.0-beta3.

Joshua-Anderson:
No problem, I thought something like that was happening.

PR #48, opened 2016-04-25
chore(git): Make builder git remote use different hostname

krancour:
This transitions the CLI from creating git `deis` git remotes that use the same hostname as the controller to using one with a hostname _derived_ from the controller's hostname.

This is to support the contingency wherein some scenario has forced an admin to use _two_ different load balancers-- one for builder and another for everything else.

See https://github.com/deis/controller/issues/657 for further context.

### DIFF
--- a/pkg/git/git_test.go
+++ b/pkg/git/git_test.go
@@ -7,15 +7,15 @@ import (
 func TestRemoteURL(t *testing.T) {
 	t.Parallel()
 
-	actual := RemoteURL("example.com", "app")
-	expected := "ssh://git@example.com:2222/app.git"
+	actual := RemoteURL("deis.example.com", "app")
+	expected := "ssh://git@deis-builder.example.com:2222/app.git"
 
 	if actual != expected {
 		t.Errorf("Expected %s, Got %s", expected, actual)
 	}
 
 	actual = RemoteURL("deis.10.245.1.3.xip.io:31350", "velcro-underdog")
-	expected = "ssh://git@deis.10.245.1.3.xip.io:2222/velcro-underdog.git"
+	expected = "ssh://git@deis-builder.10.245.1.3.xip.io:2222/velcro-underdog.git"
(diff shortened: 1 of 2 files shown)


pkg/git/git_test.go
```diff
@@ -8,14 +8,14 @@ func TestRemoteURL(t *testing.T) {
 	t.Parallel()
 
 	actual := RemoteURL("example.com", "app")
-	expected := "ssh://git@example.com:2222/app.git"
+	expected := "ssh://git@example-builder.com:2222/app.git"
```
bacongobbler: this doesn't seem right to me. So if someone wanted to expose the builder on example.com, they'd have to own both example.com and example-builder.com?

krancour: @bacongobbler, I think I mindlessly updates that test case without realizing it was at least an _edge_ case, or possibly even an _invalid_ case.

In the typical case, the controller would be at the `deis.example.com`, which, according to this logic, would make `deis-builder.example.com` the subdomain for the builder.

I'm trying to think of a scenario where the controller _isn't_ exposed as some subdomain of the platform domain.  From the router's perspective, the controller is "just another app," which is named `deis`, so it automatically occupies the `deis` subdomain of the platform domain (e.g. `example.com`).  Like other apps, it _could_ have a custom / vanity URL such as `mydeis.com`, but that's only achieved through some fairly major modifications to the charts...

I'm debating whether this edge case is worth addressing vs whether is is _so "edgy"_ that I should just update this test case to be more realistic.

Thoughts?

bacongobbler: Yeah I don't think this edge case is worth addressing since we don't route on TLDs. I think it's best to just update the unit tests to `deis.example.com` instead of `example.com` so others don't get tripped up by this either.

krancour:
@bacongobbler updated as we discussed.

PR #49, opened 2016-04-26
fix(logger): Split on the double new lines when printing log messages

jchauncey:


bacongobbler:
I just made an identical change in my client as well. `deis logs` is parsed nicely again. I'll file an issue for the butt-ugly response data that the client has to parse, though.

krancour:
@jchauncey can you just reference whatever controller change this complements?

bacongobbler:
No controller changes required. The root issue is https://github.com/deis/workflow/issues/202 so this works around that response from the controller. In my port of https://github.com/deis/controller/pull/592 from v1 the expected HTTP response was not the same as we expected.

mboersma:
```
--- FAIL: TestAppsLogs (0.01s)
	apps_test.go:370: Expected test foo bar, Got "test foo bar"
	apps_test.go:370: Expected test, Got "test"
```

bacongobbler:
@jchauncey would you mind fixing up the unit tests?

jchauncey:
yeah didnt realize they would break

PR #53, opened 2016-05-04
fix(*): fix missing error checks caught by ineffassign

Joshua-Anderson:


bacongobbler:
ping @arschles; this might have been what was causing those panics

PR #57, opened 2016-05-10
fix(apps:create): Better output when --no-remote used

### LINKED ISSUE #56: Weird UX on deis create --no-remote
This issue _might_ be with the controller and not the CLI (I have not dug far enough to know yet), but CLI is my best guess.

```
[kent@mbp ~]$ deis create --no-remote
Creating Application... done, created queens-gemstone
remote available at ssh://git@deis-builder.krancour.deis.ninja:2222/queens-gemstone.git
```

"remote available at ...?"  I asked for _no_ remote.  Indeed, no `deis` remote gets created locally _and_ I'll admit it's useful info insofar as I can infer the app name from it, but even so... it looks pretty weird.

It would probably be better to just respond saying the app named <whatever> had been created.

krancour:
Fixes #56

### DIFF
--- a/cmd/apps.go
+++ b/cmd/apps.go
@@ -57,7 +57,11 @@ func AppCreate(id string, buildpack string, remote string, noRemote bool) error 
 		}
 	}
 
-	fmt.Println("remote available at", git.RemoteURL(c.ControllerURL.Host, app.ID))
+	if noRemote {
+		fmt.Printf("If you want to add a git remote for this app later, use `deis git:remote -a %s`\n", app.ID)
+	} else {
+		fmt.Println("remote available at", git.RemoteURL(c.ControllerURL.Host, app.ID))
+	}
 
 	return nil
 }


cmd/apps.go
```diff
@@ -57,7 +57,11 @@ func AppCreate(id string, buildpack string, remote string, noRemote bool) error
 		}
 	}
 
-	fmt.Println("remote available at", git.RemoteURL(c.ControllerURL.Host, app.ID))
+	if noRemote {
+		fmt.Printf("If you want to add a git remote for this app later, use `deis git:remote -a %s`\n", app.ID)
+	} else {
+		fmt.Println("remote available at", git.RemoteURL(c.ControllerURL.Host, app.ID))
```
Joshua-Anderson: On line `56` the function returns after the remote is created, so `noRemote` will always return true at this point.

Joshua-Anderson: Sorry, my bad, I misread this :blush:. LGTM!

PR #58, opened 2016-05-11
ref(*): rewrite to use path/filepath for better OS compatibility

### LINKED ISSUE #30: Use the path/filepath package instead of path
The `path` package is not portable to operating systems that don't delimit paths with `/` (like Windows). By switching packages, the CLI becomes more portable.

Joshua-Anderson:
Fixes #30


mboersma:
Nice--this should help with Windows compatibility. ping @ogghead.

PR #60, opened 2016-05-13
fix(cmd): strip printLogs with only one newline

bacongobbler:
refs deis/workflow#202

### DIFF
--- a/cmd/apps.go
+++ b/cmd/apps.go
@@ -172,7 +172,7 @@ func AppLogs(appID string, lines int) error {
 
 // printLogs prints each log line with a color matched to its category.
 func printLogs(logs string) error {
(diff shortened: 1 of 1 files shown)


arschles:
@bacongobbler code LGTM. I noticed, however that in https://github.com/deis/workflow/issues/202#issuecomment-214840470, @jchauncey suggests that fixing the issue in the controller is an option. should it be done there _also_?

jchauncey:
Whatever we do the CLI needs fixing too

bacongobbler:
@arschles the problem isn't from the controller. The problem rests in the response from the logger which in turn causes the controller to return this in the API. I have PR'd the fix in deis/logger#79 :)

PR #63, opened 2016-05-16
ref(deis run): the API now uses exit_code instead of rc for the exit_code (ReturnCode)

helgi:


helgi:
Required for https://github.com/deis/controller/pull/738

PR #65, opened 2016-05-18
ref(cmd): clarify deis registry usage

bacongobbler:
related to deis/workflow#239. Also increases test coverage around this so a user cannot accidentally fat-finger an invalid key.

### DIFF
--- a/cmd/registry.go
+++ b/cmd/registry.go
@@ -124,8 +124,12 @@ func parseInfo(item string) (string, string, error) {
 	parts := strings.SplitN(item, "=", 2)
 
 	if len(parts) != 2 {
-		return "", "", fmt.Errorf(`%s is invalid, Must be in format key=value
+		return "", "", fmt.Errorf(`%s is invalid. Must be in format key=value
 Examples: username=bob password=s3cur3pw1`, item)
+	}
+
+	if parts[0] != "username" && parts[0] != "password" {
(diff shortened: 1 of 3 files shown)


cmd/registry.go
```diff
@@ -124,9 +124,13 @@ func parseInfo(item string) (string, string, error) {
 	parts := strings.Split(item, "=")
 
 	if len(parts) != 2 {
-		return "", "", fmt.Errorf(`%s is invalid, Must be in format key=value
+		return "", "", fmt.Errorf(`%s is invalid. Must be in format key=value
 Examples: username=bob password=s3cur3pw1`, item)
 	}
 
+	if parts[0] != "username" && parts[0] != "password" {
```
helgi: We'll just have to remember to update this when `hostname` (will be needed for Builder in the future) and potentially other items become available

parser/registry.go
```diff
@@ -61,15 +61,17 @@ Options:
 
 func registrySet(argv []string) error {
 	usage := `
-Sets registry information for an application.
-
-key/value pairs used to configure / authenticate against external registries
+Sets registry information for an application. These credentials are the same as you'd use when
```
mboersma: Could you rephrase the second sentence not to use "you"? Maybe:
> These credentials are the same as those used for `docker login` to the external registry.

bacongobbler: always with the first person, eh? Old habits die hard. Good catch.

Issue #73, opened 2016-06-06
upload builds to S3


bacongobbler:
bintray's terms of use does not allow development builds to be uploaded. Because of this, we hit https://github.com/deis/workflow/issues/282. Moving us back to S3 and updating `install-v2.sh` would help us in this regard.

Not labelling as a showstopper for v2 but certainly something we need to do ASAP.

arschles:
possibly related to https://github.com/deis/minio/issues/108, as the builds could be uploaded to the same GCS bucket..

bacongobbler:
Places so far that I've seen that need updating once we make this change include:

 - deis/deis.io (install scripts)
 - deis/workflow-e2e (fetches client binary by sha)

arschles:
A Google Cloud Storage bucket has been created in the `deis-downloads` project at https://console.cloud.google.com/storage/browser/workflow-cli/?project=deis-downloads

arschles:
See https://github.com/deis/deisrel/pull/88 for an example implementation

PR #78, opened 2016-06-09
feat(build): upload CLI binaries to Google Cloud Storage

arschles:
This change uploads binaries to GCS instead of bintray. It preserves
tag-based releases as well.

Fixes https://github.com/deis/workflow-cli/issues/73
See https://github.com/deis/workflow/issues/282#issuecomment-224008617 for additional context on why we're submitting this change.

TODO:

- [x] Fix README herein to point to Google Cloud Storage URLs
- [ ] ~~Fix external documentation (in `deis/workflow`) to point to GCS URLs~~ - created https://github.com/deis/workflow-cli/issues/79 to separately track this


Makefile
```diff
@@ -55,11 +45,12 @@ build: binary-build
 	@$(call check-static-binary,deis)
 
 build-all:
-	${DEV_ENV_CMD} gox -verbose \
-	${GO_LDFLAGS} \
-	-os="linux darwin " \
-	-arch="amd64 386" \
-	-output="$(DIST_DIR)/deis-${VERSION}-{{.OS}}-{{.Arch}}" .
+	${DEV_ENV_CMD} gox -verbose ${GO_LDFLAGS} -os="linux darwin " -arch="amd64 386" -output="$(DIST_DIR)/deis-latest-{{.OS}}-{{.Arch}}" .
```
vdice: To clarify my understanding, this and the two commands below will:

1. update/override the existing `deis-latest-{{.OS}}-{{.Arch}}` binary with this latest version 
2. throw latest version into the `${TRAVIS_TAG}` subdir <--- What if this env var is not set/does not exist?  (It will only exist if the travis build is kicked off from pushing a git tag to the repo, right?)
3. throw latest version into the `${VERSION}` subdir

As long as I'm understanding correctly, my only real question is in 2. above...

arschles: Answers inline:


1. _update/override the existing deis-latest-{{.OS}}-{{.Arch}} binary with this latest version_ - correct
2. _throw latest version into the ${TRAVIS_TAG} subdir <--- What if this env var is not set/does not exist? (It will only exist if the travis build is kicked off from pushing a git tag to the repo, right?)_ - this env var is not set, we'll fall back to just using `${VERSION}`, which is still `$(shell git rev-parse --short HEAD)`
3. _throw latest version into the ${VERSION} subdir_ - as mentioned in the previous point, the Makefile should only be doing this if `TRAVIS_TAG` isn't set. Did I get my `ifndef` blocks wrong perhaps?

vdice: oh wow @arschles this is a case of total reader error -- had completely glossed over the `ifdef/else/endif` logic.  This is completely appropriate -- apologies for the oversight!

README.md
```diff
@@ -3,27 +3,53 @@
...
 
-Install the `deis` client from [bintray](https://bintray.com/) by running
+See the appropriate sub-section for your system to download and install the latest build of this software.
 
-	$ curl -sSL http://deis.io/deis-cli/install-v2.sh | bash
+#### 64 Bit Linux
 
-The installer will fetch the latest version of the client into your current directory.
+```console
+curl https://storage.googleapis.com/workflow-cli/deis-latest-linux-amd64 > deis && chmod +x deis
```
krancour: Total nitpick... is this the conventional way to download to a file (instead of stdout) with curl? By redirecting the output? I usually use the `-o` option. Not saying one is more right than the other-- just asking the question.

README.md
```diff
@@ -3,27 +3,53 @@
...
+```
 
-	$ ./deis --version
```
krancour: Should this line move instead of going away entirely? We probably do want the users to do this in order to verify their download (and version thereof).

arschles: good point. I'll re-add

README.md
krancour: And this is non-blocking, btw.

arschles: not really sure what the conventional way is to be honest, but I'd rather use the convention. the [`install-v2.sh` script](http://deis.io/deis-cli/install-v2.sh) seems to use `-O`, so that leads me to believe that since `-o` is the closest option that does what we need here, it might be the best. Thoughts?

krancour: Good point. I think we should make this look as similar to what's in the official install script as we can. fwiw, the install script can actually lose a line (the `mv`) by using `-o` instead of `-O`.

README.md
arschles: re-added in https://github.com/deis/workflow-cli/pull/78/commits/6ca3b7771a2afc3aa7f34d49ab042e887e7c9cb0

README.md
arschles: Replaced the stdout redirection code in each command with `-o` in https://github.com/deis/workflow-cli/pull/78/commits/6cd90112f51f089cc8ca2efa43070e5ca27b0bcd. Also, I'm going to leave the `mv` command in here because I believe it makes the instructions more explicit and that's a good thing in this case.

Also, I'm going to leave the modifications to the install script out of scope of this PR, and do them as part of https://github.com/deis/workflow-cli/issues/79.

krancour:
LGTM3 @arschles

arschles:
thanks @krancour for giving this another look!

PR #80, opened 2016-06-14
test(client): clean up profile test

Joshua-Anderson:
Just cleaner code that avoids repetition.

### DIFF
--- a/controller/client/client_test.go
+++ b/controller/client/client_test.go
@@ -31,7 +31,13 @@ func createTempProfile(contents string) error {
 	return nil
 }
 
+type comparison struct {
+	key      interface{}
+	expected interface{}
+}
+
 func TestLoadSave(t *testing.T) {
+	// Load profile from file and confirm it is correctly parsed.
 	if err := createTempProfile(sFile); err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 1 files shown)


controller/client/client_test.go
```diff
@@ -90,29 +97,37 @@ func TestLoadSave(t *testing.T) {
...
+	checkComparisons(tests, t)
+}
 
-	expectedI = 100
-	if client.ResponseLimit != expectedI {
-		t.Errorf("Expected %d, Got %d", expectedI, client.ResponseLimit)
+func checkComparisons(tests []comparison, t *testing.T) {
+	for _, check := range tests {
+		if check.key != check.expected {
+			t.Errorf("Expected %v, Got %v", check.key, check.expected)
```
arschles: @Joshua-Anderson this func is great, but it would be better if the caller could call `t.Errorf` instead of here, so that the lines represented in error message reflect where the error actually occurred. thoughts on having this func not take a `*testing.T` param and return an error if `check.key != check.expected`?

Joshua-Anderson:
I'm going to close this because this is going to be part of a refactor to use the go sdk.

PR #81, opened 2016-06-14
feat(apps): print output from command to stderr if the command failed

### LINKED ISSUE #69: Failed "deis run" command should output to stderr
Currently `deis run boguscmd` sends error output to stdout. See deis/workflow-e2e#212 and [this test spec](https://github.com/deis/workflow-e2e/blob/master/tests/apps_test.go#L224-L225).

Joshua-Anderson:
Fixes #69 

Ideally we would direct individual lines to STDOUT and STDERR based on what the command itself does... however, the kubernetes api does not distinguish between the STDOUT and STDERR, so I'm not sure that is possible. I think this is probably the best we can do until we get interactive run sessions implemented.

### DIFF
--- a/cmd/apps.go
+++ b/cmd/apps.go
@@ -202,7 +202,12 @@ func AppRun(appID, command string) error {
 		return err
 	}
 
-	fmt.Print(out.Output)
+	if out.ReturnCode == 0 {
+		fmt.Print(out.Output)
+	} else {
+		fmt.Fprint(os.Stderr, out.Output)
+	}
+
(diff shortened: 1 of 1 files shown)


arschles:
@Joshua-Anderson can you put a milestone on this PR?

mboersma:
This will require an [e2e change](https://github.com/deis/workflow-e2e/blob/master/tests/apps_test.go#L224-L225) as well.

Joshua-Anderson:
See deis/workflow-e2e#231 for the e2e test fix.

PR #82, opened 2016-06-15
style(deis.go): alphabetize commands

bacongobbler:
just a small nitpick that bothered me while trying to find commands. This alphabetizes the commands such that they are easier to grok.


Joshua-Anderson:
I'm going to miss that semi-random list of commands, some in order of when they were added, others added near other relevant commands. 👻 😛 👍 

Thanks for fixing this! For some reason I completely forgot to after copying the list from the python CLI.

PR #83, opened 2016-06-15
feat(deis.go): implement `deis healthchecks` client commands

bacongobbler:
This is the client-side implementation for deis/workflow#322.

TODO:

 - [x] add deprecation notice for anyone setting HEALTHCHECK_*
 - [x] remove bacongobbler alias from glide.yaml before merging

depends on https://github.com/deis/controller-sdk-go/pull/10

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -90,6 +90,16 @@ func ConfigSet(appID string, configVars []string) error {
 		configMap["SSH_KEY"] = base64.StdEncoding.EncodeToString([]byte(sshKey))
 	}
 
+	// NOTE(bacongobbler): check if the user is using the old way to set healthchecks. If so,
+	// send them a deprecation notice.
+	for key, _ := range configMap {
+		if strings.Contains(key, "HEALTHCHECK_") {
+			fmt.Println(`Hey there! We've noticed that you're using 'deis config:set HEALTHCHECK_URL'
+to set up healthchecks. This functionality has been deprecated. In the future, please use
+'deis healthchecks' to set up application health checks. Thanks!`)
+		}
+	}
+
 	fmt.Print("Creating config... ")
 
 	quit := progress()
--- a/cmd/healthchecks.go
+++ b/cmd/healthchecks.go
@@ -1,0 +1,106 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/deis/controller-sdk-go/api"
+	"github.com/deis/controller-sdk-go/config"
+)
+
+// HealthchecksList lists an app's healthchecks.
+func HealthchecksList(appID string) error {
+	c, appID, err := load(appID)
+
+	if err != nil {
+		return err
+	}
+
+	config, err := config.List(c, appID)
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("=== %s Healthchecks\n\n", appID)
+
+	fmt.Println("--- Liveness")
+	if livenessProbe, found := config.Healthcheck["livenessProbe"]; found {
+		fmt.Println(livenessProbe)
+	} else {
+		fmt.Println("No liveness probe configured.")
+	}
+
+	fmt.Println("\n--- Readiness")
+	if readinessProbe, found := config.Healthcheck["readinessProbe"]; found {
+		fmt.Println(readinessProbe)
+	} else {
+		fmt.Println("No readiness probe configured.")
+	}
+	return nil
+}
+
+// HealthchecksSet sets an app's healthchecks.
+func HealthchecksSet(appID, healthcheckType string, probe *api.Healthcheck) error {
+	c, appID, err := load(appID)
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Applying %s healthcheck... ", healthcheckType)
+
+	quit := progress()
+	configObj := api.Config{}
+	configObj.Healthcheck = make(map[string]*api.Healthcheck)
+
+	configObj.Healthcheck[healthcheckType] = probe
+
+	_, err = config.Set(c, appID, configObj)
+
+	quit <- true
+	<-quit
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Print("done\n\n")
+
+	return HealthchecksList(appID)
+}
+
+// HealthchecksUnset removes an app's healthchecks.
+func HealthchecksUnset(appID string, healthchecks []string) error {
+	c, appID, err := load(appID)
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Print("Removing healthchecks... ")
+
+	quit := progress()
+
+	configObj := api.Config{}
+
+	healthcheckMap := make(map[string]*api.Healthcheck)
+
+	for _, healthcheck := range healthchecks {
+		healthcheckMap[healthcheck] = nil
+	}
+
+	configObj.Healthcheck = healthcheckMap
+
+	_, err = config.Set(c, appID, configObj)
+
+	quit <- true
+	<-quit
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Print("done\n\n")
+
+	return HealthchecksList(appID)
+}
--- a/parser/healthchecks.go
+++ b/parser/healthchecks.go
@@ -1,0 +1,253 @@
+package parser
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/deis/workflow-cli/cmd"
+
+	"github.com/deis/controller-sdk-go/api"
+	docopt "github.com/docopt/docopt-go"
+)
+
+// Healthchecks routes ealthcheck commands to their specific function
+func Healthchecks(argv []string) error {
+	usage := `
+Valid commands for healthchecks:
+
+healthchecks:list        list healthchecks for an app
+healthchecks:set         set healthchecks for an app
+healthchecks:unset       unset healthchecks for an app
+
+Use 'deis help [command]' to learn more.
+`
+
+	switch argv[0] {
+	case "healthchecks:list":
+		return healthchecksList(argv)
+	case "healthchecks:set":
+		return healthchecksSet(argv)
+	case "healthchecks:unset":
+		return healthchecksUnset(argv)
+	default:
+		if printHelp(argv, usage) {
+			return nil
+		}
+
+		if argv[0] == "healthchecks" {
+			argv[0] = "healthchecks:list"
+			return healthchecksList(argv)
+		}
+
+		PrintUsage()
+		return nil
+	}
+}
+
+func healthchecksList(argv []string) error {
+	usage := `
+Lists healthchecks for an application.
+
+Usage: deis healthchecks:list [options]
+
+Options:
+  -a --app=<app>
+    the uniquely identifiable name of the application.
+`
+
+	args, err := docopt.Parse(usage, argv, true, "", false, true)
+
+	if err != nil {
+		return err
+	}
+
+	return cmd.HealthchecksList(safeGetValue(args, "--app"))
+}
+
+func healthchecksSet(argv []string) error {
+	usage := `
+Sets healthchecks for an application.
+
+By default, Workflow only checks that the application starts in their Container. A health
+check may be added by configuring a health check probe for the application. The health
+checks are implemented as Kubernetes Container Probes. A 'liveness' and a 'readiness'
+probe can be configured, and each probe can be of type 'httpGet', 'exec' or 'tcpSocket'
+depending on the type of probe the Container requires.
+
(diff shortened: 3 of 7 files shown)


Joshua-Anderson:
@bacongobbler could you possibly make a PR of the stuff in the `controller` folder to [deis/controller-sdk-go](https://github.com/deis/controller-sdk-go) in addition to this? I'm pretty much done with the api extraction and will be opening a PR to transition to the SDK in a matter of a few hours.

bacongobbler:
Absolutely! I'd be happy to make that PR and migrate some of this to the SDK while this is still in flight. :)

controller/api/config.go
```diff
@@ -12,14 +12,42 @@ type ConfigUnset struct {
...
+}
+
+type ExecProbe struct {
+	Command []string `json:"command"`
+}
+
+type HTTPGetProbe struct {
+	Path        string   `json:"path"`
+	Port        string   `json:"port"`
+	HTTPHeaders []KVPair `json:"httpHeaders"`
```
kmala: i could see more options which could be specified for httpget  probe
http://kubernetes.io/kubernetes/third_party/swagger-ui/#!/api%2Fv1/createNamespacedReplicationController

helgi: I think some of those we may not want to include but I'm open to discussion

controller/api/config.go
```diff
@@ -12,14 +12,42 @@ type ConfigUnset struct {
...
+	Value string `json:"value"`
+}
+
+type ExecProbe struct {
+	Command []string `json:"command"`
+}
+
+type HTTPGetProbe struct {
+	Path        string   `json:"path"`
+	Port        string   `json:"port"`
```
helgi: I removed the support for that (at least temporarily) in the controller as @smothiki had made it so the port was pulled from the `Service` - So if you want to keep this then we need to fix up the Controller as well

controller/api/config.go
bacongobbler: Nice! Specifically the following point out at me:

 - there is a tcpSocket probe type
 - httpGet probe supports host and scheme, but not httpHeaders (probably reading old documentation)
 - periodSeconds
 - successThreshold
 - failureThreshold

controller/api/config.go
bacongobbler: @kmala is on the case for server-side implementation.

In order for this probe to work we made `port` mandatory. See deis/workflow#322 and please provide any input you have over there. This is just a reflection of that document.

controller/api/config.go
helgi: It is not old documentation:
http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_probe and http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_httpgetaction

Former should show you available types and the latter available httpget options

I found that doc page honestly much easier to read than the swagger UI :)

Just to reiterate: the Controller does not support host nor scheme and we have a want to add HttpHeaders but the others it does understand :)

bacongobbler: done. added period seconds, success/failure threshold and httpHeaders but not host and scheme. also didn't add tcpSocket probe since we don't support TCP just yet

bacongobbler:
Going to manually tell glide to install from my sdk branch so others can test (as well as jenkins)

bacongobbler:
no longer a WIP. ready for primetime!

Joshua-Anderson:
Right now this PR is still using the controller-sdk-go from your fork. I actually think this is a good thing because with my `io.Reader` PR (deis/controller-sdk-go#11), I removed verbose errors until my error handling PR lands. That PR also broke your controller-sdk-go PR.

Since the 2.1 release is coming soon, this might be a good solution until I can get error handling worked out (probably going to be tomorrow), without holding everything up.

An alternative is we create a v2.1 branch with your healthcheck pr and without my breaking changes and use that for the CLI v2.1 release.

bacongobbler:
Just re-iterating what I said on Slack:

The workflow-cli healtcheck PR is only pointing to my branch of the SDK temporarily so others can test the server-side implementation against this PR. I will be rebasing the SDK PR separately. As soon as this has 2 LGTMs, I will merge the SDK PR, then point workflow-cli back against deis/controller-sdk-go HEAD.

parser/healthchecks.go
```diff
@@ -0,0 +1,190 @@
...
+    the healthcheck type, such as 'liveness' or 'readiness'.
+  <probe-type>
+    the healthcheck probe type, such as 'httpGet', 'exec' or 'tcpSocket'.
+  <args>
+    The arguments required for the healthcheck probe.
+
+Options:
+  -a --app=<app>
+    the uniquely identifiable name for the application.
+  -p --path
```
kmala: this and all the below should be similar to --path=<path>...otherwise doopt treats them as bool and keeps False as default value

bacongobbler: good catch. fixed!

cmd/healthchecks.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Applying %s healthcheck... ", healthcheckType)
+
+	quit := progress()
+	configObj := api.Config{}
+
+	configObj.Healthcheck[healthcheckType] = probe
```
kmala: configObj.Healthcheck need to be intitialized

bacongobbler:
Okay, this is all finally ready for review.

TESTING:

1. Provision a Workflow cluster with deis/controller#836
2. build the client from source
3. start playing around with `deis healthchecks`, verifying that the expected healthcheck is set and that all I/O looks good

bacongobbler:
pinging @Joshua-Anderson as the current CI error is what caused me to open #97

bacongobbler:
turns out I was rebased on an old release of controller-sdk-go. rebasing fixed it right up

helgi:
`healtchecks` is missing when doing `deis help`

parser/healthchecks.go
```diff
@@ -0,0 +1,245 @@
...
+	return cmd.HealthchecksList(safeGetValue(args, "--app"))
+}
+
+func healthchecksSet(argv []string) error {
+	usage := `
+Sets healthchecks for an application.
+
+By default, Workflow only checks that the application starts in their Pod. A health check
+may be added by configuring a health check probe for the application. The health checks
+are implemented as Kubernetes Pod Probes. A 'liveness' and a 'readiness' probe can be
```
helgi: They are technically called `Container Probes` as k8s hasn't implemented `Pod Probes` yet

parser/healthchecks.go
```diff
@@ -0,0 +1,245 @@
...
+configured, and each probe can be of type 'httpGet' or 'exec' depending on the type of
```
helgi: `tcpSocket` not supported?

bacongobbler:
Good catch. I think I missed that when I accidentally overwrote my commit history. I'll patch that up.

parser/healthchecks.go
bacongobbler: tcpSocket is supported. This is more purgatory that was missed

parser/healthchecks.go
```diff
@@ -0,0 +1,245 @@
...
+A 'liveness' probe is useful for applications running for long periods of time, eventually
+transitioning to broken states and cannot recover except by restarting them.
+
+Other times, a 'readiness' probe is useful when the Pod is only temporarily unable to
+serve, and will recover on its own. In this case, if a Pod fails its 'readiness' probe,
+the Pod will not be shut down, but rather the Pod will stop receiving incoming requests.
+
+'httpGet' probes are just as it sounds: it performs a HTTP GET operation on the Pod. A
+response code inside the 200-399 range is considered a pass. 'httpGet' probes accept a
+port number to perform the HTTP GET operation on the Pod.
```
helgi: Should call out the path here, and make it explicit (like the docs) that only path is supported, no query params or anchor, or full URL

parser/healthchecks.go
```diff
@@ -0,0 +1,245 @@
...
+    the healthcheck type, such as 'liveness' or 'readiness'.
+  <probe-type>
+    the healthcheck probe type, such as 'httpGet', 'exec' or 'tcpSocket'.
+  <args>
+    The arguments required for the healthcheck probe.
+
+Options:
+  -a --app=<app>
+    the uniquely identifiable name for the application.
+  -p --path=<path>
```
helgi: Any reason why this and header are flags but other probes do not use that?

parser/healthchecks.go
```diff
@@ -0,0 +1,245 @@
...
+
+Usage: deis healthchecks:set <health-type> <probe-type> [options] [--] <args>...
+
+Arguments:
+  <health-type>
+    the healthcheck type, such as 'liveness' or 'readiness'.
+  <probe-type>
+    the healthcheck probe type, such as 'httpGet', 'exec' or 'tcpSocket'.
+  <args>
+    The arguments required for the healthcheck probe.
```
helgi: It feels very unclear to me what this is referencing

parser/healthchecks.go
bacongobbler: do you know of another way to set option flags for a specific argument? fwiw these options are not used when a different probe is not used. If you know of a better way to pass optional arguments I'm all ears. :)

parser/healthchecks.go
bacongobbler: This is in reference to the arguments described above, e.g. exec is a string of args, httpGet uses a port and tcpSocket accepts a port

helgi: We need to call that out more clearly, with examples and what the individual types support, probably as a list to be clearer

helgi:
`deis healthchecks:set readiness exec ls -all -a oddest-lungfish -p /healthz2 --initial-delay 30`

This command files but quoting `ls -all` works. Do we have the ability to say "give us the remainder of positional args" for `<args>` - We may be able to get past the quoting limit that way, or else we have to document that

bacongobbler:
Yes, we can supply a `--` like we do for `deis run` to work around this.

cmd/config.go
```diff
@@ -90,6 +90,14 @@ func ConfigSet(appID string, configVars []string) error {
 		configMap["SSH_KEY"] = base64.StdEncoding.EncodeToString([]byte(sshKey))
 	}
 
+	// NOTE(bacongobbler): check if the user is using the old way to set healthchecks. If so,
+	// send them a deprecation notice.
+	if _, ok := configMap["HEALTHCHECK_URL"]; ok {
```
helgi: Would it be added complication if we made this a glob?

bacongobbler: I figured that since HEALTHCHECK_URL is the trigger to start the legacy healthcheck we only need to check for that one environment variable. Do you just want to check for any signs of `HEALTHCHECK_` in the `configMap`?

helgi: Yeah, then people would be alerted if they were adjusting any part of the healthchecks instead of just url.

It's optional tho, I'm not going to push it as a required change

helgi:
@bacongobbler as long as that is called out in our documentation. And maybe even as an example in the CLI

bacongobbler:
> as long as that is called out in our documentation. And maybe even as an example in the CLI

It is called out in our documentation :)  https://github.com/deis/workflow/pull/322

bacongobbler:
Just gonna address https://github.com/deis/workflow-cli/pull/83#discussion_r68453145 and https://github.com/deis/workflow-cli/pull/83#discussion_r68450532 and I think we're all done

bacongobbler:
k, addressed. Should be good to go for review again once CI goes green.

bacongobbler:
FYI, here's a list of all the PRs across the deis org:

https://github.com/deis/controller-sdk-go/pull/10
https://github.com/deis/controller/pull/836
https://github.com/deis/workflow/pull/322

Still need e2e tests :/

helgi:
LGTM - still think the there should be usage examples and that we should ensure people know to use quoting if not using `--` but we can also just point to the documentation "for more info" kind of dealio

bacongobbler:
yeah the docs point out `--` and it's in the usage string. I don't think the client should be showing exmaples on how to use itself; that is what the official documentation is supposed to be for. :)

helgi:
Should we call out the docs themselves? per command for example

bacongobbler:
Nah if we do that we should probably add that to `deis help` in a separate PR. It'd be weird to have `deis help healthchecks` the only command that links to deis.com/docs/workflow/

helgi:
Could be both, deis help has the main doc and each command has links relevant to their own area. That can be a different PR like you said

bacongobbler:
Indeed. Not trying to shoot down your thoughts! Just thinking of the implications

helgi:
It's a very good point, consistency is good

PR #84, opened 2016-06-16
ref(*): use go sdk instead of built in code

Joshua-Anderson:
:boom:

Remove code shared with the api in the `controller` directory and use the sdk for all api interactions.


cmd/utils.go
```diff
@@ -88,3 +89,19 @@ func limitCount(objs, total int) string {
 
 	return fmt.Sprintf(" (%d of %d)\n", objs, total)
 }
+
+// checkErrors handles specific behavior for certain errors,
+// such as printing an warning for the API mismatch error
+func checkErrors(c *client.Client, err error) error {
```
bacongobbler: small nit, but mind if we call this `checkAPICompatibility` again? The old name made it more clear that this is the main intention of `checkErrors`. If we need to do more error checking in the future then we can call it `checkErrors`, imo :)

cmd/auth.go
```diff
@@ -3,35 +3,34 @@ package cmd
...
 	}
 
-	controllerURL, err := chooseScheme(*u)
+	tempClient, err := settings.Load()
 
-	if err != nil {
-		return err
+	if err == nil && tempClient.ControllerURL.Host == c.ControllerURL.Host {
+		c.Token = tempClient.Token
 	}
```
arschles: does this also need a check for `err != nil`, or is that handled in the below?

cmd/auth.go
```diff
@@ -61,19 +60,15 @@ func Register(controller string, username string, password string, email string,
...
-	}
-
 	err = auth.Register(c, username, password, email)
 
 	c.Token = ""
 
-	if err != nil {
+	fmt.Println("---")
+	fmt.Println(err)
+	fmt.Println("___")
```
arschles: should this error go to STDERR?

cmd/auth.go
Joshua-Anderson: This checks if the `*client.Client` properly loaded before trying to access `tempClient`, which otherwise could be `nil`.

Joshua-Anderson:
I ran the e2e tests locally with this PR and they passed.

cmd/auth.go
```diff
@@ -244,15 +230,15 @@ func Cancel(username string, password string, yes bool) error {
 	err = auth.Delete(c, username)
 	cleanup := fmt.Errorf("\n%s %s\n\n", "409", "Conflict")
 	if reflect.DeepEqual(err, cleanup) {
```
arschles: this could use some cleaning up. can you put an issue in the controller-sdk-go repo to have `auth.Delete` return a well-known error?

arschles: also, not blocking for this PR 😄

cmd/builds.go
```diff
@@ -50,7 +49,7 @@ func BuildsCreate(appID, image, procfile string) error {
 		if procfileMap, err = parseProcfile([]byte(procfile)); err != nil {
 			return err
 		}
-	} else if _, err := os.Stat("Procfile"); err == nil {
+	} else if _, err = os.Stat("Procfile"); err == nil {
```
arschles: nit: think you can keep the `:=`? may as well keep the error in block scope

settings/settings.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, err
+	}
+
+	settings := settingsFile{}
+	if err = json.Unmarshal(contents, &settings); err != nil {
+		return nil, err
+	}
+
+	c, err := client.New(settings.VerifySSL, settings.Controller, settings.Token,
+		settings.Username)
```
arschles: nit: can leave this on the same line or put all params on a newline?

cmd/auth.go
Joshua-Anderson: We need better errors all around the SDK! I referenced this specific code earlier in my better errors issue deis/controller-sdk-go#7

arschles:
I left a few nits, but I'm good with all this, even if you don't change any of them

Joshua-Anderson:
@arschles Fixed!

PR #101, opened 2016-06-24
fix(ps): restarting a single pod created by Deployments was not working

helgi:
Deployments created a different pod name then what RCs do (we have less flexibility): oddest-lungfish-cmd-1269180282-60byi

This way we know the proc type is always at the front of the slice, after replacing the app name


vdice:
code LGTM but can we add a unit test around this logic?

helgi:
I couldn't spot any existing unit tests

PR #103, opened 2016-06-27
ref(settings): manage username and limits in settings

Joshua-Anderson:
This complements deis/controller-sdk-go#16 by adding in a settings object to manage limits and usernames.

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -36,22 +36,23 @@ func progress() chan bool {
 	return quit
 }
 
-func load(appID string) (*deis.Client, string, error) {
-	c, err := settings.Load()
+// load loads settings file and looks up the app name
+func load(appID string) (*settings.Settings, string, error) {
+	s, err := settings.Load()
 
 	if err != nil {
 		return nil, "", err
 	}
 
 	if appID == "" {
-		appID, err = git.DetectAppName(c.ControllerURL.Host)
+		appID, err = git.DetectAppName(s.Client.ControllerURL.Host)
 
 		if err != nil {
 			return nil, "", err
 		}
 	}
 
-	return c, appID, nil
+	return s, appID, nil
 }
 
 func drinkOfChoice() string {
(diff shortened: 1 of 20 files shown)


Joshua-Anderson:
Travis CI is failing right now because of caching. When I update this to point to the `deis/controller-go-sdk` when deis/controller-sdk-go#16 is merged it will pass.

cmd/utils.go
```diff
@@ -36,22 +36,22 @@ func progress() chan bool {
 	return quit
 }
 
-func load(appID string) (*deis.Client, string, error) {
-	c, err := settings.Load()
+func load(appID string) (*deis.Client, string, int, error) {
```
bacongobbler: we should probably document this because it's unknown what `load` is intended to load up at first glance, and what exactly the values returned represent.

bacongobbler: also why don't we just return the `settings` object instead of the client and limit separately?

Joshua-Anderson:
@bacongobbler I added documentation to `load` and changed it to return a  settings object.

glide.lock
```diff
@@ -1,8 +1,10 @@
-hash: efb390ce6569c895e553582bd010493901e3fa4733fb9037970003c870cb5df2
-updated: 2016-06-27T11:56:08.014343879-07:00
+hash: a9ade02839b3b4a568225bf30592fbe944f4d74f60ec08e054d4b42f19f93271
+updated: 2016-06-27T11:39:10.012207091-07:00
 imports:
 - name: github.com/deis/controller-sdk-go
-  version: 3b6502cecedc6ebd7b0f42a8c01c31e8292d1e0d
+  version: 629d23864538f209adba0ce95304dfb6e5aab292
+  repo: https://github.com/Joshua-Anderson/controller-sdk-go.git
```
mboersma: Shouldn't this be vendoring https://github.com/deis/controller-sdk-go?

Joshua-Anderson: This PR depends on deis/controller-sdk-go#16. I didn't want to marge that PR until this one has been LGTMed. The procedure we have been using to deal with this is to temporary point the depency to the PR branch, then merge SDK PR and update the vendored dependency to point at the main repo again.

mboersma: Understood--thanks for the explanation.

PR #105, opened 2016-06-27
ref(travis): remove vendor cache

bacongobbler:
This cache directory was useful when glide was slow downloading k8s.io/kubernetes, but updates have
made fetches significantly quicker. The cache has been recently giving us more headaches with the
new SDK, so removing this should make builds more reliable again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ language: generic
 branches:
   only:
   - master
-cache:
-  directories:
-  - vendor
 services:
 - docker
 sudo: required


arschles:
Sucks that we have to do this. Is it biting us when we update dependencies?

bacongobbler:
yes, it is biting us when updating dependencies to test changes in controller-sdk-go.

bacongobbler:
as an example, my fight with https://github.com/deis/workflow-cli/pull/83 could only be resolved by clearing the cache every time I wanted to test changes

Joshua-Anderson:
An example error message is: https://travis-ci.org/deis/workflow-cli/builds/140628107#L146

arschles:
@bacongobbler @Joshua-Anderson ok. maybe worth trying to repro locally and filing an issue on the glide repo?

bacongobbler:
sure, I'll try that out and see if there's a bug in glide

Issue #61, opened 2016-05-14
Specifying an invalid limit creates a noop release


slack:
```console
$ deis limits:set web=512Mi
web=512Mi doesn't fit format type=#unit or type=#
Examples: web=2G worker=500M web=300
Applying limits... done

=== indoor-whitecap Limits

--- Memory
web     128M

--- CPU
web     40m
```

Should probably exit non-zero instead.

PR #109, opened 2016-06-30
fix(scale):dont call controller if there is no valid scale pattern

kmala:
Output before:
```
keerthanmala$ deis scale web=-1
'web=-1' does not match the pattern 'type=num', ex: web=2
Scaling processes... but first, coffee!
done in 0s
=== woodsy-jokester Processes
--- cmd:
woodsy-jokester-v2-cmd-zbonf up (v2)
```
after:
```
deis scale web=-1 -a=woodsy-jokester
'web=-1' does not match the pattern 'type=num', ex: web=2
```

### DIFF
--- a/cmd/ps.go
+++ b/cmd/ps.go
@@ -53,7 +53,7 @@ func PsScale(appID string, targets []string) error {
 				return err
 			}
 		} else {
-			fmt.Printf("'%s' does not match the pattern 'type=num', ex: web=2\n", target)
+			return fmt.Errorf("'%s' does not match the pattern 'type=num', ex: web=2\n", target)
 		}
 	}
 
--- a/cmd/ps_test.go
+++ b/cmd/ps_test.go
@@ -1,6 +1,39 @@
 package cmd
 
-import "testing"
+import (
+	"github.com/deis/workflow-cli/settings"
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestScaleFail(t *testing.T) {
+	tmpDir, err := ioutil.TempDir("", "tmpdir")
+	if err != nil {
+		t.Fatalf("error creating temp directory (%s)", err)
+	}
+	err = os.Mkdir(tmpDir+"/.deis", 0666)
+	if err != nil {
+		t.Fatalf("error creating temp directory (%s)", err)
+	}
+	os.Setenv("DEIS_PROFILE", "testing")
+	settings.SetHome(tmpDir)
+	data := []byte(`{"username":"test","ssl_verify":false,"controller":"http://deis.127.0.0.1.nip.io","token":"test","response_limit":0}`)
+	if err := ioutil.WriteFile(tmpDir+"/.deis/testing.json", data, 0644); err != nil {
+		t.Fatalf("error creating %s/.deis/testing.json (%s)", tmpDir, err)
+	}
+	defer func() {
+		if err := os.RemoveAll(tmpDir); err != nil {
+			t.Fatalf("failed to remove creds file from %s (%s)", tmpDir, err)
+		}
+	}()
+
+	expected := "'web=-1' does not match the pattern 'type=num', ex: web=2\n"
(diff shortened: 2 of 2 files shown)


cmd/ps.go
```diff
@@ -56,7 +56,9 @@ func PsScale(appID string, targets []string) error {
 			fmt.Printf("'%s' does not match the pattern 'type=num', ex: web=2\n", target)
```
bacongobbler: I think we should just make this return an error if there's a bad seed in the scale list.

kmala: i am ok with returning and error but i think we are not returning an error because user can try to scale multiple proc types at a time and only few can be bad.

bacongobbler: yeah but what I mean is that if there is a single bad scale type we should error and say "hey, this process type is poorly formed, so we're not going to accept this request."

Same thing applies for `config:set` as well. For example there were issues like https://github.com/deis/workflow-cli/issues/61 that was unintended, so there is precedence.

bacongobbler: also the current implementation will cause ps:scale to return 0 (meaning success) when they call `deis ps:scale web=-1`. That's not really what we would want to do in this scenario

bacongobbler:
In the future we should probably clean this out so we don't have to have this ugly testing code to mock out the client settings, but this works for now. Thanks!

cmd/ps_test.go
```diff
@@ -0,0 +1,35 @@
...
+	tmpDir, err := ioutil.TempDir("", "tmpdir")
+	if err != nil {
+		t.Fatalf("error creating temp directory (%s)", err)
+	}
+	err = os.Mkdir(tmpDir+"/.deis", 0666)
+	if err != nil {
+		t.Fatalf("error creating temp directory (%s)", err)
+	}
+	os.Setenv("DEIS_PROFILE", "testing")
+	os.Setenv("HOME", tmpDir)
```
Joshua-Anderson: This will break on windows, cause home is different. I wrote `settings.SetHome` to handle the cross platform differences transparently.

Joshua-Anderson:
The way I've normally tested the stuff is by extracting the logic I want to test into it's own function and then test that, which IMO is a bit cleaner. `cmd.chooseKey` is [an example of this](https://github.com/deis/workflow-cli/blob/master/cmd/keys_test.go#L133-L160).

I also have some mock session code in `settings_tests.go` that could probably be made public and does all this work for you.

bacongobbler:
that is indeed the better approach, but I personally am happy to merge this as-is and we can clean this up later.

cmd/ps.go
```diff
@@ -53,7 +53,7 @@ func PsScale(appID string, targets []string) error {
 				return err
 			}
 		} else {
-			fmt.Printf("'%s' does not match the pattern 'type=num', ex: web=2\n", target)
+			return fmt.Errorf("'%s' does not match the pattern 'type=num', ex: web=2\n", target)
```
arschles: can this become a struct that implements `error`?

bacongobbler: you mean like https://github.com/deis/controller-sdk-go/blob/master/errors.go? Those errors are for server-side, but that is a good suggestion.

arschles: not a global var, a global type since this has variable data in it:

```go
type errInvalidScalePattern struct {
  target string
}

// Error implements the error interface
func (e errInvalidScalePattern) Error() string {
  return fmt.Sprintf("'%s' does not match the pattern 'type=num', ex: web=2\n", e.target)
}
```

kmala: this is something which needs to be done on the whole repo....may be i can create an issue and do it later?

arschles: @kmala an issue to do it later is fine, but why not start with this one now?

kmala: i had created an issue https://github.com/deis/workflow-cli/issues/129

PR #110, opened 2016-07-01
bug(cmd): add more advanced handling of app urls

Joshua-Anderson:
Previously, we took a very naive approach to determining the app url. We did appID**~~deis~~**.cluster.com

This approach has a problem however, because you can assign subdomains to an app that mask another app:

```console
$ deis domains:add a -a t
$ deis apps:create a --no-remote
$ deis apps:open -a a # opens a.cluster.com, which actually points to t
```

Now, `apps:open` and `apps:info` use the app's first domain, usually the one that is assigned by default. If the app doesn't have any domains, like `a` in the example does, it'll print that the app doesn't have any domains.

This also allows me to remove `api.Apps.URL` from the SDK laster, removing the naive behavior in a way that doesn't break anything.

### DIFF
--- a/cmd/apps.go
+++ b/cmd/apps.go
@@ -11,6 +11,7 @@ import (
 	"github.com/deis/controller-sdk-go/api"
 	"github.com/deis/controller-sdk-go/apps"
 	"github.com/deis/controller-sdk-go/config"
+	"github.com/deis/controller-sdk-go/domains"
 	"github.com/deis/workflow-cli/pkg/git"
 	"github.com/deis/workflow-cli/pkg/webbrowser"
 	"github.com/deis/workflow-cli/settings"
@@ -104,11 +105,20 @@ func AppInfo(appID string) error {
 		return err
 	}
 
+	url, err := appURL(s, appID)
+	if err != nil {
+		return err
+	}
+
+	if url == "" {
+		url = fmt.Sprintf(noDomainAssignedMsg, appID)
+	}
+
 	fmt.Printf("=== %s Application\n", app.ID)
 	fmt.Println("updated: ", app.Updated)
 	fmt.Println("uuid:    ", app.UUID)
 	fmt.Println("created: ", app.Created)
-	fmt.Println("url:     ", app.URL)
+	fmt.Println("url:     ", url)
 	fmt.Println("owner:   ", app.Owner)
 	fmt.Println("id:      ", app.ID)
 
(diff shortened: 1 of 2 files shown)


bacongobbler:
I'm not sure I like this change. This makes it so `-a` accepts a domain as well as app IDs rather than strictly app IDs. That makes it inconsistent and could cause confusion among the other commands that only accept app IDs. Why don't we just do a check that app `a` exists and return a 404 if it doesn't exist instead?

Joshua-Anderson:
@bacongobbler This doesn't affect the behavior of `-a` at all... this behavior is for looking up the URL for an app when using `app:open` and `app:info`. There were a few edge cases where what the previously the *assumed* app url could actually point to another app. Now, `apps:open` and `apps:info` look up the app domain from the controller rather than try to guess it.

bacongobbler:
Ah sorry, now that I've read it more thoroughly I see what you're trying to do. :+1:

cmd/apps.go
```diff
@@ -104,11 +106,19 @@ func AppInfo(appID string) error {
...
+		}
+		url = errNoDomain.Error()
+	}
+
 	fmt.Printf("=== %s Application\n", app.ID)
 	fmt.Println("updated: ", app.Updated)
 	fmt.Println("uuid:    ", app.UUID)
 	fmt.Println("created: ", app.Created)
-	fmt.Println("url:     ", app.URL)
+	fmt.Println("url:     ", url)
```
bacongobbler: I feel like we shouldn't be changing this because the server fills out this field for us. If the server is returning us false information when the app has no domain, then we should fix this in the controller.

bacongobbler: Additionally, this adds an additional API request to this command which we try to avoid.

Joshua-Anderson: This isn't from the controller.... it's [from the SDK](https://github.com/deis/controller-sdk-go/blob/master/apps/apps.go#L108), which is making the URL assumption of appname.cluster.com

Joshua-Anderson: I intend to remove app.URL from the SDK after this is merged.

cmd/apps.go
```diff
@@ -276,3 +285,32 @@ func AppTransfer(appID, username string) error {
 
 	return nil
 }
+
+var errNoDomain = errors.New("No domain assigned to app")
```
bacongobbler: errNoDomain should say which app has no domain assigned to it

cmd/apps.go
```diff
@@ -104,11 +106,19 @@ func AppInfo(appID string) error {
 		return err
 	}
 
+	url, err := appURL(s, appID)
+	if checkAPICompatibility(s.Client, err) != nil {
```
bacongobbler: doesn't appURL already do an API compatibility check on line 294? Is there significance to checking it twice?

Joshua-Anderson:
@bacongobbler Fixed the issues you brought up!

cmd/apps.go
```diff
@@ -276,3 +284,33 @@ func AppTransfer(appID, username string) error {
 
 	return nil
 }
+
+var errNoDomain error
```
arschles: can we remove this error from package-global scope? if it needs to be package-global, then make it a `struct`

arschles:
moving out of the 2.2 milestone

Joshua-Anderson:
@arschles I refactored the code to instead check for a empty string rather than an error, removing the need to a package level error.

PR #112, opened 2016-07-05
tests(ci): switch over to using jenkins

Joshua-Anderson:


Joshua-Anderson:
This is no longer work in process and is ready to review.

A few current problems though:

1. Building and uploading SDK artifacts take 4-8 minutes, when the tests only take 2 minutes. One thing we could do is only build the `linux-amd64` binary for pull requests, since the main purpose of pull request artifacts is to use them for e2e testing. Then on master we could build the entire matrix of binaries because wait times are less of an issue.
2. One odd thing with problem 1 is that the building artifacts on my computer take ~30 seconds. On the cluster gox thinks it has a single core and bumping parallelism explicitly doesn't bring the build time time for artifacts under 4:30. Does jenkins do resource limiting on builds or something like that, because I thought those VMs have multiple cores?
3. Jenkins builds PRs using the merge commit with master that github generates, so artifacts have a different SHA than the commit a developer pushes.

Jenkinsfile
```diff
@@ -0,0 +1,88 @@
...
+parallel(
+	revision: {
+		node('linux') {
+			def gopath = pwd() + "/gopath"
+			env.GOPATH = gopath
+			def workdir = gopath + workpath_linux
+
+			dir(workdir) {
+					checkout scm
+					sh 'make bootstrap'
```
vdice: @Joshua-Anderson do you know if in this latter `parallel` section when we run a new `node('linux')` if we get a different node then what was used above?  Just wondering if we need to run `make bootstrap` again (as it is done above) -- which technically we wouldn't have to if we were using same node...

Joshua-Anderson: It can be a new node, so we have to.

Jenkinsfile
```diff
@@ -0,0 +1,88 @@
...
+			checkout scm
+		stage 'Install Windows'
+			bat pscmd('.\\make bootstrap')
+		stage 'Test Windows'
+			bat pscmd('.\\make test')
+	}
+}
+
+node('linux') {
+	def gopath = pwd() + "/gopath"
```
vdice: minor nit: what do you think about wrapping this linux gopath setup into its own closure, as it appears used identically in 3 spots?

vdice:
Aside from nit, LGTM.  Good work on this.  Do we want to set up an issue for investigating any/all of the remaining issues mentioned in https://github.com/deis/workflow-cli/pull/112#issuecomment-231124694 ?

Joshua-Anderson:
@vdice I added a closure for the gopath setup stuff. I also set CI to only build `amd64` binaries for pull requests, shaving about 4 minutes off the build time.

Jenkinsfile
```diff
@@ -0,0 +1,108 @@
...
+	},
+	latest: {
+		node('linux') {
+			def gopath = gopath_linux()
+			def workdir = workdir_linux(gopath)
+
+			dir(workdir) {
+					checkout scm
+
+					// HACK: Recommended approach for getting command output is writing to and then reading a file.
```
arschles: can you file a GH issue to fix this?

Joshua-Anderson: I could open one, but there isn't anything we can do about it. This is the way jenkins pipeline recommends doing this. Hopefully they'll come up with a better system, but that will probably be a while.

bacongobbler: we should still file an issue :)

Makefile
```diff
@@ -44,14 +54,19 @@ glideup:
...
-	${DEV_ENV_CMD} gox -verbose ${GO_LDFLAGS} -os="linux darwin windows" -arch="amd64 386" -output="$(DIST_DIR)/${TRAVIS_TAG}/deis-${TRAVIS_TAG}-{{.OS}}-{{.Arch}}" .
+build-latest:
+	${DEV_ENV_CMD} gox -verbose -parallel=3 ${GO_LDFLAGS} -os="${BUILD_OS}" -arch="${BUILD_ARCH}" -output="$(DIST_DIR)/deis-latest-{{.OS}}-{{.Arch}}" .
+
+build-revision:
+ifdef GIT_TAG
+	${DEV_ENV_CMD} gox -verbose -parallel=3 ${GO_LDFLAGS} -os="${BUILD_OS}" -arch="${BUILD_ARCH}" -output="$(DIST_DIR)/${GIT_TAG}/deis-${GIT_TAG}-{{.OS}}-{{.Arch}}" .
 else
-	${DEV_ENV_CMD} gox -verbose ${GO_LDFLAGS} -os="linux darwin windows" -arch="amd64 386" -output="$(DIST_DIR)/${VERSION}/deis-${VERSION}-{{.OS}}-{{.Arch}}" .
+	${DEV_ENV_CMD} gox -verbose -parallel=3 ${GO_LDFLAGS} -os="${BUILD_OS}" -arch="${BUILD_ARCH}" -output="$(DIST_DIR)/${VERSION}/deis-${VERSION}-{{.OS}}-{{.Arch}}" .
```
arschles: may need to turn some container CPU knobs in order to get this to build faster. my guess is increasing the cpu quota may help

arschles:
LGTM. After you merge, do you have plans to remove the appveyor and travis integrations on this repo?

Makefile
Joshua-Anderson: The builders are m3.medium, which has only a single CPU, which is why the build speeds are slow. I can increase parallelism by building on more than 2 workers, but that seemed a little excessive to me.

Joshua-Anderson:
Yep, I will disable them once I merge.

PR #121, opened 2016-07-08
[WIP] test(jenkins): trigger e2e tests after successful build

Joshua-Anderson:


Joshua-Anderson:
I added a manual input stage if e2e fails, giving you a option to retry just the e2e stage instead of the entire build in the case of a failure. Thanks @vdice for the tip! One side effect however, is that the build will remain yellow until you take manual action, not giving an obvious CI failure sign. I personally think the tradeoff is worth it, but others may disagree.

vdice:
This requires https://github.com/deis/e2e-runner/pull/27 as well, to make sure the cli commit is handed down properly to the e2e tests...

vdice:
Currently seeing job fail around the `Git Info` stage:
```
+ git describe --all --exact-match
fatal: no tag exactly matches '92959bd52a574c48b7e1d87829ad0ee60207b768'`
```
Going to branch off of this PR so I can debug and amend if needed.

vdice:
This work now exists in https://github.com/deis/workflow-cli/pull/124; closing this.

PR #124, opened 2016-07-11
test(jenkins): trigger e2e tests after successful build

vdice:
Branched from original PR https://github.com/deis/workflow-cli/pull/121


arschles:
@vdice if I'm reading this right, the e2e tests will run against CLIs that were built from PRs as well, right?

vdice:
@arschles that is correct, yes.  (you'll note just the downstream job name changes, `workflow-test` if master commit, `workflow-test-pr` if PR commit... either way, it passes down current commit for e2e to find appropriate binary)

PR #118, opened 2016-07-07
feat(cmd): add auth:whoami --all

bacongobbler:
~~requires changes in deis/controller-sdk-go before it is considered feature-complete.~~

closes #54 
requires deis/controller-sdk-go#36

PR #128, opened 2016-07-14
feat(Jenkinsfile): receive sdk-related params

vdice:
if this pipeline is triggered by a PR in controller-sdk-go

Ref https://github.com/deis/controller-sdk-go/issues/39
Ref https://github.com/deis/controller-sdk-go/pull/41

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,7 @@ def workpath_linux = "/src/github.com/deis/workflow-cli"
 def keyfile = "tmp/key.json"
 
 def getBasePath = { String filepath ->
-	def filename = filepath.lastIndexOf("/")
+	def filename = filepath.lastIndexOf(File.separator)
 	return filepath.substring(0, filename)
 }
 
@@ -23,6 +23,18 @@ def gopath_linux = {
 def workdir_linux = { String gopath ->
 	gopath + workpath_linux
 }
+
+properties([[$class: 'ParametersDefinitionProperty',
+	parameterDefinitions: [
+		[$class: 'StringParameterDefinition',
+			defaultValue: '',
+			description: 'controller-sdk-go commit sha to use in glide.yaml (not updated if empty)',
+			name : 'SDK_SHA'],
+		[$class: 'StringParameterDefinition',
+			defaultValue: '',
+			description: 'controller-sdk-go repo to use in glide.yaml (not updated if empty)',
+			name: 'SDK_GO_REPO']
+		]]])
 
 node('windows') {
 	def gopath = pwd() + "\\gopath"
@@ -95,6 +107,20 @@ parallel(
 					}
 
 					sh 'make bootstrap'
+
+					if (SDK_GO_REPO && SDK_SHA) {
+						echo "Updating local glide.yaml with controller-sdk-go repo '${SDK_GO_REPO}' and version '${SDK_SHA}'"
+
+						def pattern = "github\\.com\\/deis\\/controller-sdk-go\\n\\s+version:\\s+[a-f0-9]+"
+						def replacement = "${SDK_GO_REPO.replace("/", "\\/")}\\n  version: ${SDK_SHA}"
+						sh "perl -i -0pe 's/${pattern}/${replacement}/' glide.yaml"
+
+						def glideYaml = readFile('glide.yaml')
+						echo "Updated glide.yaml:\n${glideYaml}"
+
+						sh 'make glideup'
+					}
+
 					sh 'make build-revision'
 
 					upload_artifacts(keyfile)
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -95,6 +95,17 @@ parallel(
 					}
 
 					sh 'make bootstrap'
+
+					if (env.SDK_GO_REPO && env.SDK_SHA) {
+						echo "Updating local glide.yaml with controller-sdk-go repo '${env.SDK_GO_REPO}' and version '${env.SDK_SHA}'"
+						sh "perl -i -0pe 's/github\\.com\\/deis\\/controller-sdk-go\\n\\s+version:\\s+[a-f0-9]+/${env.SDK_GO_REPO}\\n  version: ${env.SDK_SHA}/' glide.yaml"
```
vdice: Had prev been trying the groovy route of:
```
def pattern = ~/github.com\/deis\/controller-sdk-go\n\s+version:\s+[a-f0-9]+/
def g = new File('glide.yaml')
text = g.text
g.withWriter { w ->
  w << text.replaceAll(pattern, "${env.SDK_GO_REPO}\n  version: ${env.SDK_SHA}")
}
```
Which is acceptable groovy and runs great, but the Jenkins Secure Script Plugin warns not to approve use of all/part of it:
```
signature : new java.io.File java.lang.String Approving this signature may introduce a security vulnerability! You are advised to deny it.
```
(Can be seen [here](https://ci.deis.io/scriptApproval/) as well)

vdice:
:boom: first build with successful use of PR sdk go repo & sha: https://ci.deis.io/job/Deis/job/workflow-cli/job/PR-128/39/console

bacongobbler:
This'll be awesome for testing PRs like https://github.com/deis/workflow-cli/pull/118!

vdice:
This PR should be merged before https://github.com/deis/controller-sdk-go/pull/41

Jenkinsfile
```diff
@@ -95,6 +111,20 @@ parallel(
...
+						echo "Updating local glide.yaml with controller-sdk-go repo '${SDK_GO_REPO}' and version '${SDK_SHA}'"
+
+						def pattern = "github\\.com\\/deis\\/controller-sdk-go\\n\\s+version:\\s+[a-f0-9]+"
+						def replacement = "${SDK_GO_REPO.replace("/", "\\/")}\\n  version: ${SDK_SHA}"
+						sh "perl -i -0pe 's/${pattern}/${replacement}/' glide.yaml"
+
+						def glideYaml = readFile('glide.yaml')
+						echo "Updated glide.yaml:\n${glideYaml}"
+
+						sh 'make glideup'
```
arschles: one note here - this build target will run `glide up` (obviously) which may update other dependencies, resulting in a build environment different from that which is specified in the PR...

arschles:
@vdice left a note at https://github.com/deis/workflow-cli/pull/128#discussion_r71023996, which may be an issue going forward, but I don't believe is a blocker for this particular PR. @bacongobbler since you also commented on this PR, do you have any thoughts on the issue detailed there?

Jenkinsfile
```diff
@@ -24,6 +24,22 @@ def workdir_linux = { String gopath ->
...
+			description: 'controller-sdk-go commit sha to use in glide.yaml (not updated if empty)',
+			name : 'SDK_SHA'],
+		[$class: 'StringParameterDefinition',
+			defaultValue: '',
+			description: 'controller-sdk-go repo to use in glide.yaml (not updated if empty)',
+			name: 'SDK_GO_REPO'],
+		[$class: 'BooleanParameterDefinition',
+			defaultValue: true,
+			description: 'option to run e2e tests (default: true)',
+			name: 'RUN_E2E']
```
arschles: echoing my comment at echoing my comment at https://github.com/deis/controller-sdk-go/pull/41/files#r71024488 - where do this and the above env vars come from?

vdice: @arschles they can be passed down from any upstream job that chooses to do so (for our intended case, they are passed down by the upstream `controller-sdk-go` job for PRs in the code [here](https://github.com/vdice/controller-sdk-go/blob/410d82be404ff97f72ee956bbba46e3918585f70/Jenkinsfile#L75-L83)).  If, however, no upstream job chooses to pass values for those, you can see the defaults values are empty strings (and thus the `glide.yaml` modification logic will not be executed).

That being said, I need to update this PR to remove the `RUN_E2E` -- I intend to PR this functionality separately.

vdice: now updated (`RUN_E2E` property removed)

Jenkinsfile
vdice: @arschles shall we create a ticket to find an alternative to `glide up`? Or, if one is available immediately, should we use it here?  (only update a particular package would be ideal, right?)

arschles: yea, particular package would be ideal here. I'm not really sure of a good alternative off the top of my head 😦

jchauncey: why wouldnt we just do a `glide install`?

bacongobbler: we are updating the controller-sdk-go package's version number, which then glide.lock is out of date so `glide install` won't work in this case.

PR #141, opened 2016-07-20
feat(git): add more flags to the git command

### LINKED ISSUE #39: Make the `git:remote` command print out the remote's value even if the remote already exists
Currently, the output looks like this:

```console
ENG000656:k8s-claimer aaronschlesinger$ deis git:remote
fatal: remote deis already exists.
Error: exit status 128
```

Would be nicer if it printed out what the remote should be, but it could still exit non-zero

Joshua-Anderson:
Requires workflow-e2e#278
Fixes #39

### DIFF
--- a/cmd/apps.go
+++ b/cmd/apps.go
@@ -50,17 +50,18 @@ func AppCreate(id string, buildpack string, remote string, noRemote bool) error 
 	if !noRemote {
 		if err = git.CreateRemote(s.Client.ControllerURL.Host, remote, app.ID); err != nil {
 			if err.Error() == "exit status 128" {
-				fmt.Println("To replace the existing git remote entry, run:")
-				fmt.Printf("  git remote rename deis deis.old && deis git:remote -a %s\n", app.ID)
+				msg := "A git remote with the name %s already exists. To overwrite this remote run:\n"
+				msg += "deis git:remote --force --remote %s --app %s"
+				return fmt.Errorf(msg, remote, remote, app.ID)
 			}
 			return err
 		}
+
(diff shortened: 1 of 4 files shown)


Joshua-Anderson:
This CI is going to fail because it requires deis/workflow-e2e#278, which has passes using the binaries from this PR.

vdice:
@Joshua-Anderson I see what looks like new scenarios/logic added in this PR that aren't covered in the required workflow-e2e PR (https://github.com/deis/workflow-e2e/pull/278).  Can we wrap any added logic under test?  (Or is it already covered?)

Joshua-Anderson:
I was thinking that since they were cli-only, they didn't really belong in workflow-e2e. If we were to implement component specific functional tests, they would belong there.

I was unsure of whether to add the logic to e2e or just pass on it for now. I'm happy to either.

vdice:
@Joshua-Anderson I'd say, if the anticipated timeline for introducing functional tests is a ways out, we create ticket(s) to add minimal specs in workflow-e2e covering logic added here.  Otherwise, perhaps a note can be added to the initial ticket(s) around the functional testing initiative?  All this is assuming unit tests don't apply here...

Barring all of that, should this PR be manually tested?

Joshua-Anderson:
I could write unit tests, but they would be tricky because nearly all of the PR deals with calling the git cli and reading the output.

It seems like a good approach for now could be to write e2e tests and then if we end up writing functions tests remove them.

Joshua-Anderson:
I added some basic e2e tests to cover the `git:remote` functionality.

PR #143, opened 2016-07-20
ref(makefile): remove GO15VENDOREXPERIMENT as it is no longer necessary

Joshua-Anderson:
The CLI is now built with go1.6 and docker-go-dev is using v1.6, so I don't think there's any reason to keep it around.


Joshua-Anderson:
This passed CI [here](https://ci.deis.io/job/workflow-test-pr/4180/console) but uploading test results to jenkins failed resulting in a fail.

PR #144, opened 2016-07-20
feat(deis): Sort output of different listings

nathansamson:
This will sort the output of the listing of apps, domains and
process types.

Same as deis/deis#5072

### DIFF
--- a/cmd/apps.go
+++ b/cmd/apps.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 	"time"
 
@@ -85,8 +86,14 @@ func AppsList(results int) error {
 
 	fmt.Printf("=== Apps%s", limitCount(len(apps), count))
 
+	var appNames []string
 	for _, app := range apps {
-		fmt.Println(app.ID)
+		appNames = append(appNames, app.ID)
+	}
+	sort.Strings(appNames)
+
+	for _, appName := range appNames {
+		fmt.Println(appName)
 	}
 	return nil
 }
--- a/cmd/domains.go
+++ b/cmd/domains.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/deis/controller-sdk-go/domains"
 )
@@ -25,8 +26,14 @@ func DomainsList(appID string, results int) error {
 
 	fmt.Printf("=== %s Domains%s", appID, limitCount(len(domains), count))
 
+	var domainNames []string
 	for _, domain := range domains {
-		fmt.Println(domain.Domain)
+		domainNames = append(domainNames, domain.Domain)
+	}
+	sort.Strings(domainNames)
+
+	for _, domainName := range domainNames {
+		fmt.Println(domainName)
 	}
 	return nil
 }
--- a/cmd/ps.go
+++ b/cmd/ps.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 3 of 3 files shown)


cmd/ps.go
```diff
@@ -120,10 +121,16 @@ func printProcesses(appID string, processes []api.Pods) {
...
+		psTypes = append(psTypes, psType)
+	}
+	sort.Strings(psTypes)
+
+	for _, psType := range psTypes {
 		fmt.Printf("--- %s:\n", psType)
 
-		for _, proc := range procs {
+		for _, proc := range psMap[psType] {
 			fmt.Printf("%s %s (%s)\n", proc.Name, proc.State, proc.Release)
```
Joshua-Anderson: It might make sense to sort the individual processes in addition to sorting the process types.

nathansamson: are you talking about web.1 web.2 and so on? I am pretty sure they are sorted anyways

nathansamson: Just realized this is V2 (which I am not particularly familiar with; and in fact I did not test this code with a real cluster), and I believe I saw that V2 has "cryptic" names for instances which indeed might not be sorted....

I just took the same code that I changed for V1 version (which I tested) and copied over to V2, but in fact behaviour might indeed be different

Joshua-Anderson: Yeah, workflow uses the kubernetes pod IDs for process names, unlike the ps.# format of deis v1.

Joshua-Anderson:
Thanks for the PR, I love consistant ordering!

This hit most of the major `list` commands, but it is still missing a few (off the top of my head `users:list` and `perms:list`).

If you don't want to do those, I think it can be merged as is and I can add those in later.

Also, don't worry too much about CI failures. The e2e testing is a little flakey right now, so it often requires some retries before passing. I'll keep retrying it until it passes for you.

nathansamson:
Very much possible I missed a few. I only did the ones I generally use

I'll add users and perms as well (and any others I might find)

nathansamson:
Updated the V1 client with some other listings. After that one is accepted I can port those to this branch as well

cmd/apps.go
```diff
@@ -85,8 +86,14 @@ func AppsList(results int) error {
 
 	fmt.Printf("=== Apps%s", limitCount(len(apps), count))
 
+	var appNames []string
```
bacongobbler: Instead of converting them into a slice of strings with their names, why don't we provide a new type called `Apps` in the SDK and provide the following functions to the [Apps struct](https://github.com/deis/controller-sdk-go/blob/cd8ecc1f17db726e57586a4d149b60a25f3c4dc7/api/apps.go#L3-L11)?

```
type Apps []App

func (a Apps) Len() int           { return len(a) }
func (a Apps) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
func (a Apps) Less(i, j int) bool { return a[i].ID < a[j].ID }
```

That way we can make `apps.List` return type `Apps`, which can then use `sort.Sort(apps)` without having to perform all this logic ourselves to generate a sort-able list.

Either way LGTM, just explaining a "best practices" approach to this problem. :)

cmd/domains.go
```diff
@@ -25,8 +26,14 @@ func DomainsList(appID string, results int) error {
 
 	fmt.Printf("=== %s Domains%s", appID, limitCount(len(domains), count))
 
+	var domainNames []string
```
bacongobbler: Same comment here as [above](https://github.com/deis/workflow-cli/pull/144/files#r72125711). Introducing a `Domains` object into the SDK with the applicable sorting functions required would be the optimal approach, IMO.

cmd/ps.go
```diff
@@ -120,10 +121,16 @@ func printProcesses(appID string, processes []api.Pods) {
 
 	fmt.Printf("=== %s Processes\n", appID)
 
-	for psType, procs := range psMap {
+	var psTypes []string
```
bacongobbler: Same comment here as [above](https://github.com/deis/workflow-cli/pull/144/files#r72125711). Introducing a `Processes` object into the SDK with the applicable sorting functions required would be the optimal approach, IMO.

cmd/apps.go
nathansamson: Aha did not know you could do that in Go.

I have to say I found my method a bit cumbersome but I did not know how to write a new Sort function. I did not realise it was as simple as implementing a simple (implicit) interface.

Its on my todo list to learn Go properly btw, just did not have the time or use case yet :)

nathansamson: I will update my pull request, but first at https://github.com/deis/deis/pull/5072

Do you want seperate commits for each action or can I just squash everything together at the end?

bacongobbler: One of the examples at https://golang.org/pkg/sort/ has a great example on how to do this: https://play.golang.org/p/ntNak4vvQM :)

> Do you want seperate commits for each action or can I just squash everything together at the end?

Since this kind of enhancement would require a PR at deis/controller-sdk-go for v2, one commit here is fine. Same for the v1 client, which does not require an external library.

nathansamson:
WIll close this pull request. Will forward-port once deis/deis#5072 gets accepted

PR #145, opened 2016-07-21
feat(cmd): add auth:whoami --all

### LINKED ISSUE #54: implement auth:whoami
see deis/controller#543

bacongobbler:
closes #54 
supercedes #118

### DIFF
--- a/cmd/auth.go
+++ b/cmd/auth.go
@@ -247,15 +247,24 @@ func Cancel(username string, password string, yes bool) error {
 	return nil
 }
 
-// Whoami prints the logged in user.
-func Whoami() error {
+// Whoami prints the logged in user. If all is true, it fetches info from the controller to know
+// more about the user.
+func Whoami(all bool) error {
 	s, err := settings.Load()
 
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


bacongobbler:
Ah, I think the reason why it's failing to build is because my PR was out of date with master and had an older Jenkinsfile. That should do it!

bacongobbler:
Hmm, nope that didn't do the trick. @vdice am I missing something here to make the PR build with deis/controller-sdk-go#36?

vdice:
@bacongobbler We don't yet have the mechanism in place for 'Requires' between workflow-cli and controller-sdk-go.  As of this writing, a PR in controller-sdk-go is run thru e2e by using that version in an otherwise 'master' version of workflow-cli, and, likewise, a PR in workflow-cli just defaults to using the 'master' version of controller-sdk-go.  

That being said, we should be in a good position to add the 'Requires' functionality for any PRs in the former case (controller-sdk-go).  We should be able to clone into a forked workflow-cli repo at a certain commit before building the binary with the controller-sdk-go commit.  I will create this ticket.

bacongobbler:
Adding both LGTM labels as SDK, controller and e2e `auth:whoami` PRs have been approved for merge, so I'm assuming this is fine as well.

PR #148, opened 2016-07-22
feat(bumpversion): bump cli version to 2.3.0-dev

smothiki:
bump client version to next release

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,7 +1,7 @@
 package version
 
 // Version identifies this Deis product revision.
-const Version = "2.2.0"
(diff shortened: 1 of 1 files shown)


Joshua-Anderson:
We may want to hold on this, because we are moving to releasing all components separately. Are we still going to want to bump to a development release afterwards in that case? @jchauncey What do you think?

jchauncey:
i think the cli could be an independently releasable component too. although it might be more directly tied to controller and controller-sdk changes.

smothiki:
Closing this PR as per new release process

PR #149, opened 2016-07-25
feat(shortcuts): add shortcuts test

sgoings:
- fixes deis/workflow-e2e#243
- fix various gofmt errors


Joshua-Anderson:
Sorry about the gofmt stuff, it should be less of a problem in the future after #135 is merged and it actually fails tests.

PR #151, opened 2016-07-25
fix(tests): set owner of files created by containers to the local user

Joshua-Anderson:
This fixes with jenkins failing to delete files when cleaning up.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,11 +6,20 @@ def getBasePath = { String filepath ->
 	return filepath.substring(0, filename)
 }
 
+def make = { String target ->
+	try {
+		sh "make ${target} fileperms"
+	} catch(error) {
+		sh "make fileperms"
+		false
+	}
+}
+
 def upload_artifacts = { String filepath ->
 	withCredentials([[$class: 'FileBinding', credentialsId: 'e80fd033-dd76-4d96-be79-6c272726fb82', variable: 'GCSKEY']]) {
 		sh "mkdir -p ${getBasePath(filepath)}"
 		sh "cat \"\${GCSKEY}\" > ${filepath}"
-		sh "make upload-gcs"
+		make 'upload-gcs'
 	}
 }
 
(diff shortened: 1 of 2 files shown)


Joshua-Anderson:
@vdice This seems to be failing CI due to permission issues. Any idea what the problem is?

Joshua-Anderson:
I'm changing strategy here. Instead of trying to create files with jenkins's specific UID and GID, chown the created files to the local user.

Joshua-Anderson:
@mboersma @vdice Could someone please double check on the jenkins nodes that the fileperms have been successfully set?

Joshua-Anderson:
I'm going to unset the LGTMs because I complete redid the code.

vdice:
file perms looking good on Jenkins (uid/gid of user jenkins used):
```
/home/jenkins/workspace/Deis/workflow-cli$ ls -haltr PR-151/gopath/src/github.com/deis/workflow-cli/vendor/
total 20K
drwxr-xr-x  8 jenkins jenkins 4.0K Jul 26 18:31 github.com
drwxr-xr-x  3 jenkins jenkins 4.0K Jul 26 18:31 golang.org
drwxr-xr-x  3 jenkins jenkins 4.0K Jul 26 18:31 gopkg.in
drwxr-xr-x  5 jenkins jenkins 4.0K Jul 26 18:31 .
drwxr-xr-x 13 jenkins jenkins 4.0K Jul 26 21:31 ..
```

Jenkinsfile
```diff
@@ -6,11 +6,20 @@ def getBasePath = { String filepath ->
 	return filepath.substring(0, filename)
 }
 
+def make = { String target ->
+	try {
```
vdice: Curious if this could be rewritten/simplified to:
```
def make = { String target ->
	try {
		sh "make ${target}"
	} catch(error) {
		false
	}
	sh "make fileperms"
}
```
... or would that last `sh` invocation not get run?

vdice:
curiosity note/question; otherwise LGTM

Jenkinsfile
Joshua-Anderson: false kills the builds, so it wouldn't run

Joshua-Anderson:
For some reason the status links to pipeline build #7, which failed, but the most recent build is [pipeline build #8](https://ci.deis.io/job/Deis/job/workflow-cli/job/PR-151/8/) which passed.

PR #156, opened 2016-07-29
feat(cmd): add deis routing

bacongobbler:
This command enables or disables the application being exposed external to the cluster through the
router.

see https://github.com/deis/controller/pull/934

SDK changes: https://github.com/deis/controller-sdk-go/compare/09f8465bf5598f84987734082d22f4201632a320...cd0f563b127736127e0e9debeddcbcb014621b04

### DIFF
--- a/parser/routing.go
+++ b/parser/routing.go
@@ -1,0 +1,100 @@
+package parser
+
+import (
+	"github.com/deis/workflow-cli/cmd"
+	docopt "github.com/docopt/docopt-go"
+)
+
+// Routing displays all relevant commands for `deis routing`.
+func Routing(argv []string) error {
+	usage := `
+Valid commands for routing:
+
+routing:info       view routability of an application
+routing:enable     enable routing for an app
+routing:disable    disable routing for an app
+
+Use 'deis help [command]' to learn more.
+`
+
+	switch argv[0] {
+	case "routing:info":
+		return routingInfo(argv)
+	case "routing:enable":
+		return routingEnable(argv)
+	case "routing:disable":
+		return routingDisable(argv)
+	default:
+		if printHelp(argv, usage) {
+			return nil
+		}
+
+		if argv[0] == "routing" {
+			argv[0] = "routing:info"
+			return routingInfo(argv)
+		}
+
+		PrintUsage()
+		return nil
+	}
+}
+
+func routingInfo(argv []string) error {
+	usage := `
+Prints info about the current application's routability.
+
+Usage: deis routing:info [options]
+
+Options:
+  -a --app=<app>
+    the uniquely identifiable name for the application.
+`
+
+	args, err := docopt.Parse(usage, argv, true, "", false, true)
+
+	if err != nil {
+		return err
+	}
+
+	return cmd.RoutingInfo(safeGetValue(args, "--app"))
+}
+
+func routingEnable(argv []string) error {
+	usage := `
+Enables routability for an app.
+
+Usage: deis routing:enable [options]
+
+Options:
+  -a --app=<app>
+    the uniquely identifiable name of the application.
+`
+
+	args, err := docopt.Parse(usage, argv, true, "", false, true)
+
+	if err != nil {
+		return err
+	}
+
+	return cmd.RoutingEnable(safeGetValue(args, "--app"))
+}
+
+func routingDisable(argv []string) error {
+	usage := `
+Disables routability for an app.
+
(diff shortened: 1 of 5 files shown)


Joshua-Anderson:
Would it also be possible (maybe not this sprint) to add support for toggling routable?

`POST /v2/apps/<app>` already exists for updating the owner, so maybe the routable app field can be supported there too? Then you can have a cli command like `deis apps:routing enable/disable`

Joshua-Anderson:
Could the routable field be added to `deis apps:info`?

bacongobbler:
> Would it also be possible (maybe not this sprint) to add support for toggling routable?

It could be, but what would the use case be for? Once an application is deployed, I would assume that you would not want to make it routable/unroutable. Putting a live application currently undergoing some construction into maintenance mode, however, makes sense.

jchauncey:
I can totally see apps moving from routable to not routable. Refactoring services from being public facing to living behind a firewall happens all the time.

bacongobbler:
Okay, technically yes it is possible but not required for this sprint.

bacongobbler:
For reference this whole feature is based off [this use case](https://groups.google.com/forum/#!topic/deis-users/F8G6rrDsmEk), so my assumptions are:

 - the user needs some form of service discovery between apps (good ol' k8s and docs)
 - the app must be either public or private on creation

Feature flags like `deis apps:routing enable/disable` is out of scope for this sprint, but please feel free to open a ticket on that!

jchauncey:
But doesnt the `apps:routing` configuration follow how we configure other parts of an application (like healthchecks, ports, etc)

bacongobbler:
Nope. All it does is enable/disable the `router.deis.io/routable` annotation on the service. Nothing more.

bacongobbler:
Revisted this, so now it's through the Config object so it's toggle-able. New commands are:

```
deis routing:list
deis routing:enable
deis routing:disable
```

parser/routing.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/deis/workflow-cli/cmd"
+	docopt "github.com/docopt/docopt-go"
+)
+
+// Routing displays all relevant commands for `deis routing`.
+func Routing(argv []string) error {
+	usage := `
+Valid commands for routing:
+
+routing:list       list routability for an app
```
Joshua-Anderson: I find list a `list` a little confusing, because it implies a list. Would `info` or `get` possibly make more sense?

parser/config.go
```diff
@@ -30,6 +31,8 @@ Use 'deis help [command]' to learn more.
 		return configPull(argv)
 	case "config:push":
 		return configPush(argv)
+	case "config:routing":
```
Joshua-Anderson: Do we want to support two different methods of toggling routing? I think it might make sense not to have this at all to avoid confusion.

parser/routing.go
helgi: It's a bit confusing, true, but `list` is how we do other configuration based things. Maybe it should have no `list` and only show up in the `app:info` page? Not sure... UX thing

parser/config.go
helgi: How do you mean? @bacongobbler did away with `--routing`

parser/routing.go
bacongobbler: `routing:info` sounds good to me.

Joshua-Anderson: I think just `apps:info` would be fine, but if I was going to look up routability and had no idea how, I would run `deis routing help` and would expect some singular command like `info` and would be confused if `routing:list` was what I was looking for.

parser/config.go
bacongobbler: this is old code, but it's `routing:enable` and `routing:disable` to toggle.

Joshua-Anderson: Yeah, I was confused by the existence of both `config:routing` and `routing:enable`. But if that's just old code no big deal.

bacongobbler:
Looks like we're gonna need unit tests for this to pass codecov. I'll take a look after I'm done debugging some stuff on deis v1

Joshua-Anderson:
@bacongobbler Don't worry about it. Consider the codecov status optional right now 😄 . Adding tests looks impossible until the CLI testing refactor I'm working on is complete and api interface can be mocked. I'll make sure to add tests to it along with everything else.

bacongobbler:
okay then, merging!

PR #157, opened 2016-07-29
chore(glide): update controller-sdk-go

mboersma:
Brings in saner APIMismatch behavior and an updated error message. All credit goes to @Joshua-Anderson.

Requires https://github.com/deis/controller#943


helgi:
https://github.com/deis/controller/pull/943 passed (Travis didn't but that's unrelated, some new slowness going on)

PR #160, opened 2016-08-02
feat(healthcheck): Add support for healthcheck per proctype

kmala:


cmd/healthchecks.go
```diff
@@ -2,13 +2,31 @@ package cmd
...
+	fmt.Fprintln(out, "--- Liveness")
+	if livenessProbe, found := healthcheck["livenessProbe"]; found {
+		fmt.Fprintln(out, livenessProbe)
+	} else {
+		fmt.Fprintln(out, "No liveness probe configured.")
+	}
+
+	fmt.Fprintln(out, "\n--- Readiness")
+	if readinessProbe, found := healthcheck["readinessProbe"]; found {
+		fmt.Fprintln(out, readinessProbe)
```
bacongobbler: We should display what process type this healthcheck is for.

cmd/healthchecks.go
```diff
@@ -22,25 +40,25 @@ func HealthchecksList(appID string) error {
 	}
 
 	fmt.Printf("=== %s Healthchecks\n\n", appID)
-
-	fmt.Println("--- Liveness")
-	if livenessProbe, found := config.Healthcheck["livenessProbe"]; found {
-		fmt.Println(livenessProbe)
+	if procType == "" {
+		for proc, healthcheck := range config.Healthcheck {
+			fmt.Println(proc + ":")
```
bacongobbler: ah I see it's done here. nvm

cmd/healthchecks.go
```diff
@@ -22,25 +40,25 @@ func HealthchecksList(appID string) error {
...
+			printHealthCheck(os.Stdout, *healthcheck)
+		}
 	} else {
-		fmt.Println("No liveness probe configured.")
+		fmt.Println(procType + ":")
+		if healthcheck, found := config.Healthcheck[procType]; found {
+			printHealthCheck(os.Stdout, *healthcheck)
+		} else {
+			fmt.Println(appID + " doesn't have proctype" + procType)
```
bacongobbler: we should exit 1 here

cmd/healthchecks.go
```diff
@@ -50,10 +68,12 @@ func HealthchecksSet(appID, healthcheckType string, probe *api.Healthcheck) erro
 	fmt.Printf("Applying %s healthcheck... ", healthcheckType)
 
 	quit := progress()
-	configObj := api.Config{}
-	configObj.Healthcheck = make(map[string]*api.Healthcheck)
 
-	configObj.Healthcheck[healthcheckType] = probe
+	healthcheckMap := make(api.Healthchecks)
+	healthcheckMap[healthcheckType] = probe
+	configObj := api.Config{}
```
bacongobbler: you can compress this.

```
configObj := api.Config{Healthcheck: make(map[string]*api.Healthchecks)}
```

parser/healthchecks.go
```diff
@@ -11,6 +11,10 @@ import (
 	docopt "github.com/docopt/docopt-go"
 )
 
+const (
+	defaultProcType string = "web/cmd"
```
bacongobbler: can you explain this logic a bit more? I don't see this logic explained anywhere in the docstring. I thought it had to be a single process type.

bacongobbler:
Another thought that I can't put into comments: we should be able to set the same healthcheck for multiple processes in one command (is this what the `web/cmd` syntax accomplishes?)

PR #163, opened 2016-08-03
ref(version): update the version of the client

kmala:


Joshua-Anderson:
Note that in the future I would like to see the entire version overridden by the makefile, but LGTM as a quick fix.

PR #166, opened 2016-08-05
feat(logging): extract logging to its own package and write tests

Joshua-Anderson:


pkg/logging/log_windows.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"fmt"
+	"io"
+)
+
+// PrintLog prints each log line. Windows doesn't support ansi escape codes,
+// so color is not printed.
+func PrintLog(out io.Writer, log string) error {
+	fmt.Fprintln(out, log)
```
bacongobbler: needs a `return` statement here

PR #167, opened 2016-08-05
fix(version): move all version logic to git

Joshua-Anderson:
If commit is not a tagged commit: `<latest tag>-<sha>`.
If commit is tagged `<tag>`
If version is overridden with `$VERSION`, `<version>`

cc @vdice

This way we no longer have to manually bump the version in code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,18 @@ else
 	GOOS=linux
 endif
 
+# The latest git tag on branch
+GIT_TAG := $(shell git describe --abbrev=0 --tags)
+# If the latest commit is tagged
+TAGGED_COMMIT := $(shell git tag -l --contains HEAD)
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -8,6 +8,18 @@ else
 	GOOS=linux
 endif
 
+# The latest git tag on branch
+GIT_TAG := $(shell git describe --abbrev=0 --tags)
+# If the latest commit is tagged
+TAGGED_COMMIT := $(shell git tag -l --contains HEAD)
+REVISION := $(shell git rev-parse --short HEAD)
```
vdice: should this be `REVISION ?= ...` so that a user's env can supply `REVISION`, if set?

Joshua-Anderson:
@vdice Fixed!

PR #169, opened 2016-08-09
feat(*): allow configuration flag

### LINKED ISSUE #28: The CLI should accept any path provided by DEIS_PROFILE
Right now the json files can only exists in the `.deis` directory.

Joshua-Anderson:
Usage: `deis -c path/to/config.json`

If you want to quickly switch between different clusters, it also works like $DEIS_PROFILE: `deis -c dev`

`$DEIS_PROFILE` behaves identically to the config flag.

Fixes #28

### DIFF
--- a/cmd/apps.go
+++ b/cmd/apps.go
@@ -17,8 +17,8 @@ import (
 )
 
 // AppCreate creates an app.
-func AppCreate(id string, buildpack string, remote string, noRemote bool) error {
-	s, err := settings.Load()
+func (d DeisCmd) AppCreate(id, buildpack, remote string, noRemote bool) error {
+	s, err := settings.Load(d.ConfigFile)
 	if err != nil {
 		return err
 	}
@@ -68,8 +68,8 @@ func AppCreate(id string, buildpack string, remote string, noRemote bool) error 
 }
 
 // AppsList lists apps on the Deis controller.
-func AppsList(results int) error {
-	s, err := settings.Load()
+func (d DeisCmd) AppsList(results int) error {
+	s, err := settings.Load(d.ConfigFile)
 
 	if err != nil {
 		return err
@@ -93,8 +93,8 @@ func AppsList(results int) error {
 }
 
 // AppInfo prints info about app.
-func AppInfo(appID string) error {
-	s, appID, err := load(appID)
+func (d DeisCmd) AppInfo(appID string) error {
+	s, appID, err := load(d.ConfigFile, appID)
 
 	if err != nil {
 		return err
@@ -124,13 +124,13 @@ func AppInfo(appID string) error {
 
 	fmt.Println()
 	// print the app processes
-	if err = PsList(app.ID, defaultLimit); err != nil {
+	if err = d.PsList(app.ID, defaultLimit); err != nil {
 		return err
 	}
 
 	fmt.Println()
 	// print the app domains
-	if err = DomainsList(app.ID, defaultLimit); err != nil {
+	if err = d.DomainsList(app.ID, defaultLimit); err != nil {
 		return err
 	}
 
@@ -140,8 +140,8 @@ func AppInfo(appID string) error {
 }
 
 // AppOpen opens an app in the default webbrowser.
-func AppOpen(appID string) error {
-	s, appID, err := load(appID)
+func (d DeisCmd) AppOpen(appID string) error {
+	s, appID, err := load(d.ConfigFile, appID)
 
 	if err != nil {
 		return err
@@ -164,8 +164,8 @@ func AppOpen(appID string) error {
 }
 
 // AppLogs returns the logs from an app.
-func AppLogs(appID string, lines int) error {
-	s, appID, err := load(appID)
+func (d DeisCmd) AppLogs(appID string, lines int) error {
+	s, appID, err := load(d.ConfigFile, appID)
 
 	if err != nil {
 		return err
@@ -184,8 +184,8 @@ func AppLogs(appID string, lines int) error {
 }
 
 // AppRun runs a one time command in the app.
-func AppRun(appID, command string) error {
-	s, appID, err := load(appID)
+func (d DeisCmd) AppRun(appID, command string) error {
+	s, appID, err := load(d.ConfigFile, appID)
 
 	if err != nil {
 		return err
@@ -209,10 +209,10 @@ func AppRun(appID, command string) error {
 }
 
 // AppDestroy destroys an app.
-func AppDestroy(appID, confirm string) error {
+func (d DeisCmd) AppDestroy(appID, confirm string) error {
 	gitSession := false
 
-	s, err := settings.Load()
+	s, err := settings.Load(d.ConfigFile)
 
 	if err != nil {
 		return err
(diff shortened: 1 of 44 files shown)


vdice:
@Joshua-Anderson Is there a corresponding docs PR for adding to/updating http://docs-v2.readthedocs.io/en/latest/users/cli/#support-for-multiple-profiles?

Joshua-Anderson:
E2E PR: deis/workflow-e2e#301

Joshua-Anderson:
I added the following line when logging in: `Configuration file written to /Users/<user>/.deis/client.json`. This should give people better visibility about where their sensitive files are stored and makes it easier to debug issues with setting `$DEIS_PROFILE`.

I also opened a docs PR at deis/workflow#440

vdice:
Tested locally/manually, looks good!
```
vaughndice in workflow-e2e on (master) $ deis apps:list -c ~/vrd2-config.json
=== Apps
foo
vaughndice in workflow-e2e on (master) $ DEIS_PROFILE=~/vrd-config.json deis apps:list
=== Apps
bar
```

deis.go
```diff
@@ -29,6 +29,9 @@ Option flags::
 
   -h --help     display help information
   -v --version  display client version
+  -c --config   (optional) path to configuration file. Equilivent to
```
bacongobbler: equivalent*

parser/utils.go
```diff
@@ -49,3 +49,10 @@ func printHelp(argv []string, usage string) bool {
 
 	return false
 }
+
+func addGlobalFlags(input string) string {
```
bacongobbler: global client flags don't have to be part of every help doc. It just needs to be part of the main deis usage as seen in deis.go. Why do we need it in every command?

settings/settings.go
```diff
@@ -34,12 +34,13 @@ type Settings struct {
...
 // Load loads a new client from a settings file.
-func Load() (*Settings, error) {
-	filename := locateSettingsFile()
+func Load(cf string) (*Settings, error) {
+	filename := locateSettingsFile(cf)
 
 	if _, err := os.Stat(filename); err != nil {
 		if os.IsNotExist(err) {
-			return nil, errors.New("Not logged in. Use 'deis login' or 'deis register' to get started.")
+			return nil, fmt.Errorf(`Client configuration file not found at: %s
```
bacongobbler: I'm pretty sure this will break e2e. https://github.com/deis/workflow-e2e/blob/a31e7e659d90bfb253263b16f0b3d210fc643641/tests/auth_test.go#L33

parser/utils.go
Joshua-Anderson: An alternative approach would be to parse it from the global section and pass it as an argument. In that case however, it would not show up anywhere except `deis help`, so I though it wouldn't have enough visibility.

cmd/apps.go
```diff
@@ -17,8 +17,8 @@ import (
 )
 
 // AppCreate creates an app.
-func AppCreate(id string, buildpack string, remote string, noRemote bool) error {
-	s, err := settings.Load()
+func AppCreate(cf, id, buildpack, remote string, noRemote bool) error {
+	s, err := settings.Load(cf)
```
bacongobbler: We should have a settings struct that accepts a client file location and has a Load() function. That way we don't have to pass around a string to the client file on every client command.

parser/utils.go
bacongobbler: I think that's more than enough visibility for a global flag. 👍

settings/settings.go
Joshua-Anderson: Already opened an e2e PR for it.

Joshua-Anderson:
@bacongobbler I edited the App commands to use an approach like what you suggested. What do you think?

deis.go
```diff
@@ -25,10 +26,15 @@ The Deis command-line client issues API calls to a Deis controller.
...
-
-  -h --help     display help information
-  -v --version  display client version
+Options:
+  -h --help
+    display help information
+  -v --version
+    display client version
+  -c --config=<config>
+    path to configuration file. Equilivent to
```
bacongobbler: Equivalent*

deis.go
```diff
@@ -25,10 +26,15 @@ The Deis command-line client issues API calls to a Deis controller.
...
+    setting $DEIS_PROFILE. Defaults to ~/.deis/config.json.
```
bacongobbler: isn't the default path `~/.deis/client.json`?

deis.go
```diff
@@ -161,6 +172,34 @@ Use 'git push deis master' to deploy to an application.
 	return 0
 }
 
+func removeConfigFlag(argv []string) []string {
```
bacongobbler: we should make this into a generic function in the future, but nicely done 👍

deis.go
```diff
@@ -161,6 +172,34 @@ Use 'git push deis master' to deploy to an application.
...
+	var kept []string
+	for i, arg := range argv {
+		if arg == "-c" || strings.HasPrefix(arg, "--config=") {
+			continue
+			// If the previous optoin -c, remove the arguement as well
```
bacongobbler: typo: option

bacongobbler: typo: argument

bacongobbler:
implementation looks fine, but it seems like we treat `apps` differently than the rest of the `parser` package, and we're still adding the --config flag to every other command. Is there still something blocking us here?

Joshua-Anderson:
@bacongobbler I made the changes you suggested for the Apps commands as a proof of concept to show you before doing the rest. If Apps looks good to you I'll go ahead and do the rest.

cmd/apps.go
```diff
@@ -16,9 +16,26 @@ import (
 	"github.com/deis/workflow-cli/settings"
 )
 
+// AppCmd is interface definition for running app commands
+type AppCmd interface {
```
bacongobbler: maybe once you roll this out to the rest of `cmd` you might want to rename this to `Commander`. I imagine we'd want a single interface that defines every command in the `cmd` package for mocking purposes so we don't have to create specific structs for every subcommand.

Joshua-Anderson: `Commander` I like it! 😃

I was thinking that have a few smaller interfaces might be easier to work with so I could have several smaller mocks  rather than 1 larger one. It looks like it would be the same amount of code either way, so I erred on the side of smaller interfaces.

What do you think?

bacongobbler: I think it's the same amount of code in `cmd`, but in deis.go we'd need to have an instance of every struct available there rather than one massive struct that implements the entire `cmd` package. Either way works for me though, just a heads up

Joshua-Anderson:
@bacongobbler: Ready for a review. I've updated the entire codebase to use these changes.

bacongobbler:
there's still two comments pending, otherwise LGTM:

https://github.com/deis/workflow-cli/pull/169#discussion_r74610237
https://github.com/deis/workflow-cli/pull/169#discussion_r74610436

Joshua-Anderson:
@bacongobbler Sorry I overlooked those, I fixed them up.

Joshua-Anderson:
Just rebased and added flags to `deis maintenance:*`

Joshua-Anderson:
Successful e2e run in deis/workflow-e2e#301

PR #171, opened 2016-08-11
feat(maintenance): Add support for maintenance mode for apps

kmala:


smothiki:
Does it need glide yaml file update ?? cause there is a corresponding change in go-controller-sdk

kmala:
https://github.com/deis/controller-sdk-go/pull/77 corresponding sdk change

Issue #174, opened 2016-08-12
[ci] wipe out workspace if glide install fails


vdice:
It would be handy to have logic to wipe out the Jenkins workspace (minimally, deleting the `vendor` dir) if the glide install step fails (`make bootstrap`) so we don't have to clean the slave manually.

PR #176, opened 2016-08-12
feat(Jenkinsfile): wipe vendor dir if bootstrap fails

vdice:
Closes https://github.com/deis/workflow-cli/issues/174

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,6 @@
 def workpath_linux = "/src/github.com/deis/workflow-cli"
+def windows = 'windows'
+def linux = 'linux'
 
 def getBasePath = { String filepath ->
 	def filename = filepath.lastIndexOf(File.separator)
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -54,7 +56,21 @@ def sh = { String cmd ->
...
 
-node('windows') {
+def bootstrap = { String node ->
+	bootstrapCmd = node == windows ? bat(pscmd('.\\make bootstrap')) : make('bootstrap')
+
+	try {
+		bootstrapCmd
+	} catch(error) {
+		echo "bootstrap failed; wiping 'vendor' directory and trying again..."
+		retry(1) {
```
Joshua-Anderson: Is retry smart enough not to trigger an infinite loop?

vdice: Good inquiry to check @Joshua-Anderson -- it appears `retry` does the right thing and only attempts the given number of times; see https://ci.deis.io/job/Deis/job/workflow-cli/job/PR-176/6/console

PR #181, opened 2016-08-15
feat(CI): refactor CI to build and then pass around a test image

Joshua-Anderson:
cc @jchauncey


Jenkinsfile
```diff
@@ -100,97 +47,141 @@ node(windows) {
...
 stage 'Build and Upload Artifacts'
 
+def gcs_bucket = "gs://workflow-cli"
+
+def upload_artifacts = { boolean cache ->
+	headers  = "-h 'x-goog-meta-git-branch:${git_branch}' "
+			headers += "-h 'x-goog-meta-git-sha:${git_commit}' "
+			headers += "-h 'x-goog-meta-ci-job:${env.JOB_NAME}' "
+			headers += "-h 'x-goog-meta-ci-number:${env.BUILD_NUMBER}' "
+			headers += "-h 'x-goog-meta-ci-url:${env.BUILD_URL}'"
```
helgi: So many extra tabs, intentional?

Joshua-Anderson:
This is now working and ready for reviews.

Joshua-Anderson:
How this works is that jenkins builds a docker image with the source code and scripts for running tests and linting. This image is then pulled and the specified test, lint, build, and upload commands are run.

I could not make scripts for uploading (the script is coded into the jenkinsfile) to protect GCS keys.

Jenkinsfile
```diff
@@ -100,97 +47,151 @@ node(windows) {
...
 }
 
-stage 'Git Info'
+def test_image = "quay.io/deisci/workflow-cli-dev:${git_commit.take(7)}"
+def mutable_image = 'quay.io/deisci/workflow-cli-dev:latest'
+
 node(linux) {
+		stage 'Build and push test container'
+			checkout scm
+			env.REVISION = git_commit.take(7)
```
vdice: is this `env.REVISION` used below?

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM quay.io/deis/go-dev:0.16.0
```
helgi: 0.17 is out now

Jenkinsfile
Joshua-Anderson: It was used in an earlier iteration of the PR, but not anymore, I'll remove it.

Jenkinsfile
```diff
@@ -100,97 +47,150 @@ node(windows) {
...
+			flags = ""
+			if (git_branch != "remotes/origin/master") {
+				echo "Skipping build of 386 binaries to shorten CI for Pull Requests"
+				flags += "-e BUILD_ARCH=amd64"
+			}
 
-					make 'bootstrap'
-					env.REVISION = git_commit.take(7)
-					make 'build-revision'
+			tmp_dir = mktmp()
```
vdice: nit: maybe consider containing this logic (until the `rm -rf ${tmp_dir}`) into a method?  It looks like the only difference between here and if `git_branch == "remotes/origin/master"` is the cache bool?

Joshua-Anderson:
Due to the build image update, I had to make some really minor code changes to get CI to pass.

Makefile
```diff
@@ -55,14 +55,16 @@ glideup:
 build: binary-build
 	@$(call check-static-binary,deis)
 
+# This is supposed to be run within a docker container
 build-latest:
-	${DEV_ENV_CMD} gox -verbose -parallel=3 ${GO_LDFLAGS} -os="${BUILD_OS}" -arch="${BUILD_ARCH}" -output="$(DIST_DIR)/deis-latest-{{.OS}}-{{.Arch}}" .
+	gox -verbose -parallel=3 ${GO_LDFLAGS} -os="${BUILD_OS}" -arch="${BUILD_ARCH}" -output="$(DIST_DIR)/deis-latest-{{.OS}}-{{.Arch}}" .
```
vdice: Shouldn't `$(DIST_DIR)` be `${DIST_DIR}` ?  Meaning, not treated as output from an executable but simply a var to interpolate?

Joshua-Anderson: I'll trust your judgement on it. I didn't write this code, but that sounds reasonable.

vdice:
This seems like a neat approach to container-izing the source/deps.  It will be interesting to see how we can use this approach in other Workflow component golang projects, if only to avoid installing golang deps directly on ci hosts...

Joshua-Anderson:
I optimized the `Dockerfile` to better cache vendored dependencies.

Joshua-Anderson:
I *finally* figured out the problem with @vdice's help. For some reason with go1.7 `-s` no longer seems to work. `-extldflags "-static"` seems to work though. I kept testing the linux executable on the same container that build the executable, so the dynamic libraries were all found. 

Curse the mysterious `bash: /usr/local/bin/deis file or folder not found` error message.

PR #183, opened 2016-08-17
fix(tests): start adding unit tests

Joshua-Anderson:
This adds POC unit tests for the domains part of the cmd package.

### DIFF
--- a/cmd/domains_test.go
+++ b/cmd/domains_test.go
@@ -1,0 +1,101 @@
+package cmd
+
+import (
+	"bytes"
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/arschles/assert"
+	"github.com/deis/controller-sdk-go/api"
+	"github.com/deis/workflow-cli/pkg/testutil"
+)
+
+func TestDomainsList(t *testing.T) {
+	t.Parallel()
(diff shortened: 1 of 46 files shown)


cmd/domains_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/arschles/assert"
+	"github.com/deis/controller-sdk-go/api"
+	"github.com/deis/workflow-cli/pkg/testutil"
+)
+
+func TestDomainsList(t *testing.T) {
+	t.Parallel()
```
mboersma: `t.Parallel()` always makes me 😄

cmd/keys_test.go
```diff
@@ -114,7 +114,7 @@ func TestListKeys(t *testing.T) {
 		}
 	}
 
-	keys, err := listKeys()
+	keys, err := listKeys(ioutil.Discard)
```
mboersma: Nifty--I didn't know about `ioutil.Discard`.

mboersma:
Good stuff--the new domains tests and test_utils look solid, and everything else appears to be refactoring that was required by that. 💚

PR #190, opened 2016-08-22
feat(CI): upload CLI to seperate buckets

Joshua-Anderson:
To avoid making changes, I'm going to start by uploading to both the `workflow-cli` bucket and then specific bucket. We can then start to migrate.

@vdice how do you propose looking up a revision for a CLI between the PR and master bucket in e2e? Try to download from master, and if that fails download from the PR bucket?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -122,9 +122,11 @@ parallel(
 
 stage 'Build and Upload Artifacts'
 
-def gcs_bucket = "gs://workflow-cli"
+def old_gcs_bucket = "gs://workflow-cli"
+def pr_gcs_bucket = "gs://workflow-cli-pr"
+def master_gcs_bucket = "gs://workflow-cli-master"
 
-def upload_artifacts = { String dist_dir, boolean cache ->
+def upload_artifacts = { String dist_dir, String auth_id, String bucket, boolean cache ->
 	headers  = "-h 'x-goog-meta-git-branch:${git_branch}' "
 	headers += "-h 'x-goog-meta-git-sha:${git_commit}' "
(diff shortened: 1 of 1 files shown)


bacongobbler:
is there an issue open to explain why this change is necessary? My concern is how this will affect the usability of install-v2.sh to install specific versions of the client

bacongobbler:
fwiw I'm all for this just wanna know some background :)

Joshua-Anderson:
@bacongobbler there's a few reasons for the change.

1. security: We have credentials injected into PR builds, so making sure there is no way for a PR to touch/overwrite the offical binaries is nice. A edge case in our new versioning system accidentally overwrote the v2.4.0 artifacts, but a malicious attack could do far more harm.
2. Cleanness: We want to keep the master and release binaries indefinitely, but many PR binaries are only used for CI. This makes it much easier to just clean up PR artifacts every once in a while.

> My concern is how this will affect the usability of install-v2.sh to install specific versions of the client

We'll mirror the release binaries from the `workflow-cli` bucket to the `workflow-cli-release` bucket and then change the url in the install script. It should be transparent to users. If anyone uses the GCS URL directly, we can leave the workflow-cli bucket unchanged for a while.

vdice:
> @vdice how do you propose looking up a revision for a CLI between the PR and master bucket in e2e? Try to download from master, and if that fails download from the PR bucket?

@Joshua-Anderson that does sound like a first approach.  Otherwise, if we can send down a flag (`IS_PR` or some such) from the initial Jenkins test job all the way down to the workflow-e2e container which could be used to determine the bucket to attempt pulling from.

PR #191, opened 2016-08-22
feat(cmd): add `deis tls`

bacongobbler:


cmd/tls_test.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer server.Close()
+	var b bytes.Buffer
+	cmdr := DeisCmd{WOut: &b, ConfigFile: cf}
+
+	server.Mux.HandleFunc("/v2/apps/numenor/tls/", func(w http.ResponseWriter, r *http.Request) {
+		testutil.SetHeaders(w)
+		fmt.Fprintf(w, `{
```
Joshua-Anderson: If you're not actually using the `tls` object in the CLI, I'd make the tests simpler and just return `{}`.

Joshua-Anderson: Sorry, this is the wrong spot. I ment this for enabling and disabling tests.

cmd/tls_test.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer server.Close()
+	var b bytes.Buffer
+	cmdr := DeisCmd{WOut: &b, ConfigFile: cf}
+
+	server.Mux.HandleFunc("/v2/apps/numenor/tls/", func(w http.ResponseWriter, r *http.Request) {
+		testutil.SetHeaders(w)
+		body, err := ioutil.ReadAll(r.Body)
```
Joshua-Anderson: Why are you asserting that the body is empty?

As I can see from the SDK, the body is definitely not empty.

Also, instead of doing all this work of reading the body, you can use the helper I wrote `testutil.AssertBody(t, []byte{}, r)`

cmd/tls_test.go
bacongobbler: I left it in as it is the expected response from the controller. Saves time for future changes.

cmd/tls_test.go
bacongobbler: I'm asserting that the tests included data in the body so that we can fail sooner and identify the issue.

Once this is rebased after https://github.com/deis/controller-sdk-go/pull/92 is merged, I will use the helper function. Thanks

bacongobbler: fixed

Joshua-Anderson: You can replace lines 56-61 with `testutil.AssertBody(t, api.NewTLS(), r)`. The helper means that you no longer have to use exact json for the comparisons.

bacongobbler: done.

cmd/tls_test.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer server.Close()
+	var b bytes.Buffer
+	cmdr := DeisCmd{WOut: &b, ConfigFile: cf}
+
+	server.Mux.HandleFunc("/v2/apps/numenor/tls/", func(w http.ResponseWriter, r *http.Request) {
+		testutil.SetHeaders(w)
+		testutil.AssertBody(t, []byte(`{"https_enabled": true}`), r)
```
Joshua-Anderson: It's preferred to use the struct rather than raw json so that changing the object later (or reordering fields) doesn't break in the future.

bacongobbler: The reason I did not go with that function call is because we'd have to do the following: https://github.com/deis/controller-sdk-go/blob/master/tls/tls_test.go#L171-L179

`api.NewTLS()` returns a TLS struct with HTTPSEnabled set to `false`. I went with the lazy approach to check that the response body is what we'd expect rather than set up a TLS object and test against that. Less code overall.

If you'd like me to implement the change then I'll change it, but the end result is still the same: we are asserting that the response body returns a bytestring with `{"https_enabled": true}` as the content.

Joshua-Anderson: Yeah, I figured that might be why. We're using structs everywhere else, so I'd prefer to use a struct here too, but I don't feel that strongly so if you'd prefer not to that's ok.

PR #192, opened 2016-08-22
tests(config): add test for config commands

Joshua-Anderson:
I'm skipping `config:push` and `config:pull` at this moment, because they're going to more complicated since they depend on pipes `>` and `<` for input and output.

### DIFF
--- a/cmd/config_test.go
+++ b/cmd/config_test.go
@@ -1,9 +1,14 @@
 package cmd
 
 import (
+	"bytes"
+	"fmt"
+	"net/http"
 	"testing"
 
 	"github.com/arschles/assert"
+	"github.com/deis/controller-sdk-go/api"
+	"github.com/deis/workflow-cli/pkg/testutil"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/config_test.go
```diff
@@ -16,3 +21,189 @@ func TestParseConfig(t *testing.T) {
...
+        "TEST":  "testing",
+        "NCC":   "1701",
+        "TRUE":  "false",
+        "FLOAT": "12.34"
+    },
+    "memory": {},
+    "cpu": {},
+    "tags": {},
+    "registry": {},
+    "routable": true,
```
helgi: This is no longer part of the config object

Joshua-Anderson:
@helgi Fixed!

vdice:
>I'm skipping config:push and config:pull at this moment, because they're going to more complicated since they depend on pipes > and < for input and output.

@Joshua-Anderson could you create a ticket for this work, if it hasn't already been created?

Otherwise, LGTM!

PR #193, opened 2016-08-23
chore(Makefile): update to go-dev 0.17.0

mboersma:
Includes the go 1.7 toolchain. See https://tip.golang.org/doc/go1.7.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ endif
 BUILD_OS ?=linux darwin windows
 BUILD_ARCH ?=amd64 386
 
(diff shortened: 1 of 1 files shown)


Joshua-Anderson:
I'm not sure how much sense this makes, we're only really using `DEV_ENV` for building a binary image, and we should probably transition to using the dockerfile we use for tests and linting. The dockerfile is already based of the latest `go-dev` image.

mboersma:
> The dockerfile is already based of the latest go-dev image.

I saw that, but apparently submitted this PR anyway. 🚅

PR #198, opened 2016-08-24
feat(whitelist): Add support for ip whitlising for app

kmala:


cmd/whitelist.go
```diff
@@ -0,0 +1,72 @@
...
+
+	quit := progress(d.WOut)
+	_, err = whitelist.Add(s.Client, appID, strings.Split(IPs, ","))
+	quit <- true
+	<-quit
+	if checkAPICompatibility(s.Client, err, d.WErr) != nil {
+		return err
+	}
+
+	d.Print("done\n")
```
Joshua-Anderson: I'd do `d.Println("done")` instead.

kmala: changed.

Joshua-Anderson:
Do we want to add the whitelist to `apps:info`?

It's not a blocker or anything, just a random idea.

kmala:
@Joshua-Anderson i think it will good to add but then we may also have many other which we can add so i will leave the final decision on the UX to @slack or @helgi

PR #199, opened 2016-08-24
fix(git): properly log errors from git

Joshua-Anderson:
Log the command and the Stderr that results in an git error:

```
$ deis git:remote -a test
Error: Error when running 'git remote add deis ssh://git@deis-builder.1.1.1.1.nip.io:2222/test.git'
fatal: Not a git repository (or any of the parent directories): .git
```

### DIFF
--- a/cmd/git.go
+++ b/cmd/git.go
@@ -18,7 +18,10 @@ func (d DeisCmd) GitRemote(appID, remote string, force bool) error {
 	if err != nil {
 		//If git remote doesn't exist, create it without issue
 		if err == git.ErrRemoteNotFound {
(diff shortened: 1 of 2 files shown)


Joshua-Anderson:
I unset all the LGTMs because I greatly changed the PR. Same idea, but I'm not adding the error checking to every function and added more error checks in the `cmdr.GitRemote` method as well.

Joshua-Anderson:
For some reason this is pointing to CI build 3, but build 4 passed https://ci.deis.io/job/Deis/job/workflow-cli/job/PR-199/4/.

PR #203, opened 2016-08-25
ref(Jenkinsfile): remove shell output hack

vdice:
To use approach seen [here](https://github.com/jenkinsci/pipeline-examples/blob/master/pipeline-examples/gitcommit/gitcommit.groovy).


vdice:
It looks like we have the latest versions of all of the Pipeline plugins, so trying this out.

Joshua-Anderson:
Looks like it works, nice find!

PR #205, opened 2016-08-26
feat(users): show admins when listing users

Joshua-Anderson:
Also added tests to cover the behavior

This will probably break e2e, but I'm questioning the need for `users:list` e2e tests. It's tested in the controller, the sdk, and cli, so as long as the api contract stays the same (which we commit to by using semver) it should work.

### DIFF
--- a/cmd/users.go
+++ b/cmd/users.go
@@ -22,9 +22,12 @@ func (d DeisCmd) UsersList(results int) error {
 		return err
 	}
 
-	d.Printf("=== Users%s", limitCount(len(users), count))
+	d.Printf("=== Users (*=admin)%s", limitCount(len(users), count))
 
 	for _, user := range users {
+		if user.IsSuperuser {
+			d.Print("*")
+		}
 		d.Println(user.Username)
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


cmd/users.go
```diff
@@ -22,9 +22,12 @@ func (d DeisCmd) UsersList(results int) error {
 		return err
 	}
 
-	d.Printf("=== Users%s", limitCount(len(users), count))
+	d.Printf("=== Users (* admin)%s", limitCount(len(users), count))
```
bacongobbler: `* admin` doesn't really explain that names prefixed with * means they're an admin user. Would you mind elaborating on this instead in the docstring in parser/users.go?

Joshua-Anderson: Yeah, that's why I added `=== Users (* admin)`

I can also add it to the docstring as well though.

bacongobbler: Yeah, I'm just saying that it didn't make sense to me on first glance. Maybe `(* == admin)`?

Joshua-Anderson:
@bacongobbler How does it look now?

bacongobbler:
> How does it look now?

worthy of an LGTM. 👍

PR #208, opened 2016-08-29
feat(autoscale): add the ability to define autoscale rules per process type on an app

helgi:
ref https://github.com/deis/controller-sdk-go/pull/89

### DIFF
--- a/cmd/autoscale.go
+++ b/cmd/autoscale.go
@@ -1,0 +1,91 @@
+package cmd
+
+import (
+	"github.com/deis/controller-sdk-go/api"
+	"github.com/deis/controller-sdk-go/appsettings"
+)
+
+// AutoscaleList tells the informations about app's autoscale status
+func (d DeisCmd) AutoscaleList(appID string) error {
+	s, appID, err := load(d.ConfigFile, appID)
+
+	if err != nil {
+		return err
+	}
+
+	appSettings, err := appsettings.List(s.Client, appID)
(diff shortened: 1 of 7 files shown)


cmd/autoscale_test.go
```diff
@@ -0,0 +1,106 @@
...
+			"created": "2014-01-01T00:00:00UTC",
+			"updated": "2014-01-01T00:00:00UTC",
+			"uuid": "de1bf5b5-4a72-4f94-a10c-d2a3741cdf75"
+		}`)
+	})
+
+	err = cmdr.AutoscaleList("rivendell")
+	assert.NoErr(t, err)
+	// TODO
+	assert.Equal(t, b.String(), "Maintenance mode is on.\n", "output")
```
kmala: is `maintenance mode` by mistake?

helgi: yeah, hadn't fully finished that part - Hence TODO, wanted to show you the other parts

cmd/autoscale.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	d.Printf("=== %s Autoscale\n\n", appID)
+
+	if appSettings.Autoscale == nil {
+		d.Println("No autoscale rules found.")
+	} else {
+		for process, kv := range appSettings.Autoscale {
+			d.Println("--- " + process + ":")
+			fmt.Fprintln(os.Stdout, *kv)
```
kmala: i think this should just be d.Println(*kv) instead of fmt.Fprintln()

PR #210, opened 2016-08-31
fix(ps): use new sdk for sorted processes

Joshua-Anderson:
Since iterating over a hash is random, it resulted in random ordering and occasionally broke tests.

### DIFF
--- a/cmd/ps.go
+++ b/cmd/ps.go
@@ -103,16 +103,16 @@ func (d DeisCmd) PsRestart(appID, target string) error {
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


bacongobbler:
Are any changes required in the builder? I assume not since IIRC it just uses the release object, buildhook and keyhook.

Joshua-Anderson:
@bacongobbler Nope, the builder only uses hooks. It also has a pinned SDK version, so we don't have to upgrade the SDK version in the builder unless we want to.

PR #216, opened 2016-09-02
fix(limits): remove short cpu flag

Joshua-Anderson:
It conflicts with the global short config flag

### DIFF
--- a/parser/limits.go
+++ b/parser/limits.go
@@ -91,7 +91,7 @@ Arguments:
 Options:
   -a --app=<app>
     the uniquely identifiable name for the application.
-  -c --cpu
+  --cpu
     limits CPU.
   -m --memory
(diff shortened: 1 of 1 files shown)


helgi:
Make sure to update any docs as well

helgi:
also I think we should ditch the memory short flag if CPU is getting it removed

Joshua-Anderson:
> also I think we should ditch the memory short flag if CPU is getting it removed

I like short flags, but I can see how it would get confusing.

Joshua-Anderson:
@helgi Actually, I think I'm okay with leaving `-m` because `--memory` is pretty long and annoying to type. What do you think?

helgi:
We can, I just feel like that's inconsistency is all

PR #221, opened 2016-09-06
fix(CI): explicitly set codecov git SHA

Joshua-Anderson:


Jenkinsfile
```diff
@@ -110,7 +110,16 @@ parallel(
 			withCredentials([[$class: 'StringBinding',
 												credentialsId: '995d99a7-466b-4beb-bf75-f3ba91cbbc18',
 												variable: 'CODECOV_TOKEN']]) {
-				sh "docker run -e CODECOV_TOKEN=\${CODECOV_TOKEN} --rm ${test_image} sh -c 'test-cover.sh && curl -s https://codecov.io/bash | bash'"
+				def codecov = "codecov -Z -C ${git_commit} "
+				if (git_branch == "remotes/origin/master") {
+					codecov += "-B master"
```
vdice: Just curious; running `codecov --help` I see that the `-B/-C/-P/-b/-T` vars are `"automatically detected by popular CI providers"` -- maybe some or all of these might be auto-detected?  (Though perhaps the whole point of this PR was they weren't/were incorrect!)

Joshua-Anderson: Sorry, I should have added some more context 😄

Codecov used to use the info given by jenkins environmental variables. After switching to containers, it used git to get some information, but it didn't report on all PRs due to the SHA of GH merge commits being different. Then, I excluded the `.git` directory from the container to slim down on size (It was over 20 MB), and that entirely broke codecov.

Issue #196, opened 2016-08-24
deis git:remote just prints "Exit status 129"


bkoski:
Not sure what's wrong.  I'm able to manually add the git remote.  But if I attempt to use the built-in deis command to add, I get an error 129 without much explanation:

```
deis git:remote -a MY_APP_NAME -f
Error: exit status 129
```

Joshua-Anderson:
@bk What operating system are you using? I've seen this on debian jessie but that's it. 

The error is coming from git returning a `129`.
I plan to make a PR to fix Sterr printing, so we at least get the error message soon.

bkoski:
I'm on OS X 10.10.5, `git version 2.5.4 (Apple Git-61)`, deis cli `v2.4.0`.

Joshua-Anderson:
Interesting, I don't have this problem with OSX `10.11.6` and `git version 2.7.4 (Apple Git-66)`.

I'll make a PR to log the git Stderr in the next few hours so that we can start debugging.

Joshua-Anderson:
I created PR #199 to add more git output in case of errors.

The PR binary for OS X is https://storage.googleapis.com/workflow-cli-pr/07637bd/deis-07637bd-darwin-amd64

If you're have the time and would like to try and investigate this, run the same command you did before with this binary and hopefully enough information should be logged that we can work from there.

bkoski:
Strangely, even this build returns 129 for me:

```
./deis-07637bd-darwin-amd64 git:remote -a MY_APP_NAME
Error: exit status 129
```

As far as I can tell, all of the other CLI commands are working just fine.

Joshua-Anderson:
Thanks for trying it out! I just added the better error output to `git remote add`, but I realized I should add the error output to every git command. If you have a chance, this
https://storage.googleapis.com/workflow-cli-pr/83abaa4/deis-83abaa4-darwin-amd64
should print the Stderr from git. 

Also your specific error leads me to think that `git remote get-url <remote name>` is where the error is coming from.

Joshua-Anderson:
@mboersma has confirmed the the problematic command is `git get-url`, which apparently is new in git 2.7.

I switched to parsing `git remove -v` instead in #227, which should fix the problem.

PR #222, opened 2016-09-07
fix(cmd): fix help string when git remote already exists

bacongobbler:
TODO:

 - [x] add tests

### DIFF
--- a/cmd/apps.go
+++ b/cmd/apps.go
@@ -50,7 +50,7 @@ func (d DeisCmd) AppCreate(id, buildpack, remote string, noRemote bool) error {
 
 	if !noRemote {
 		if err = git.CreateRemote(s.Client.ControllerURL.Host, remote, app.ID); err != nil {
-			if err.Error() == "exit status 128" {
+			if strings.Contains(err.Error(), fmt.Sprintf("fatal: remote %s already exists.", remote)) {
 				msg := "A git remote with the name %s already exists. To overwrite this remote run:\n"
 				msg += "deis git:remote --force --remote %s --app %s"
 				return fmt.Errorf(msg, remote, remote, app.ID)
--- a/cmd/apps_test.go
+++ b/cmd/apps_test.go
@@ -1,6 +1,18 @@
 package cmd
 
-import "testing"
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/arschles/assert"
+
+	"github.com/deis/workflow-cli/pkg/git"
+	"github.com/deis/workflow-cli/pkg/testutil"
+)
 
 type expandURLCases struct {
 	Input    string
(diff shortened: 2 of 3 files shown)


cmd/apps.go
```diff
@@ -50,7 +50,7 @@ func (d DeisCmd) AppCreate(id, buildpack, remote string, noRemote bool) error {
 
 	if !noRemote {
 		if err = git.CreateRemote(s.Client.ControllerURL.Host, remote, app.ID); err != nil {
-			if err.Error() == "exit status 128" {
+			if strings.Contains(err.Error(), "Error when running 'git remote add") {
```
Joshua-Anderson: This works, but it'd prefer to check for `fatal: remote <remote name> already exists.`, which is more explicit.

bacongobbler: Good idea. This actually caught a bug in the unit tests I'm writing too. It was catching `[Error when running 'git remote add deis ssh://git@127-builder.0.0.1:2222/foo.git'
       		fatal: Not a git repository (or any of the parent directories): .git` which is of course an error we did not intend to catch.

cmd/apps_test.go
```diff
@@ -27,3 +37,35 @@ func TestExpandUrl(t *testing.T) {
...
+    "owner": "jkirk",
+    "id": "foo",
+    "created": "2014-01-01T00:00:00UTC",
+    "updated": "2014-01-01T00:00:00UTC",
+    "uuid": "de1bf5b5-4a72-4f94-a10c-d2a3741cdf75"
+}`)
+	})
+
+	// create a remote first before running apps:create
+	assert.NoErr(t, git.Init())
```
Joshua-Anderson: Can move the PWD to a temporary directory, so we're not creating lots of git remotes in the workflow-cli directory?

bacongobbler: We run the tests in a container so it's wiped after each run. Are we to assume that these tests can be run out of the container?

Joshua-Anderson: Yeah, I like running `go test` out of a container so I can get my <1 second test runs 😄

bacongobbler: done.

cmd/apps.go
```diff
@@ -50,7 +50,7 @@ func (d DeisCmd) AppCreate(id, buildpack, remote string, noRemote bool) error {
 
 	if !noRemote {
 		if err = git.CreateRemote(s.Client.ControllerURL.Host, remote, app.ID); err != nil {
-			if err.Error() == "exit status 128" {
+			if strings.Contains(err.Error(), fmt.Sprintf("fatal: remote %s already exists.", remote)) {
```
arschles: @bacongobbler can you compare the error type here, instead of the error string?

pkg/git/git.go
```diff
@@ -34,6 +34,17 @@ func CreateRemote(host, remote, appID string) error {
 	return nil
 }
 
+// Init creates a new git repository in the local directory.
+func Init() error {
+	cmd := []string{"git", "init"}
+	_, err := exec.Command(cmd[0], cmd[1]).Output()
+	if err != nil {
```
arschles: combine this and the above line: `if _, err := exec.Command(...); err != nil {`

cmd/apps.go
bacongobbler: Unfortunately all git errors fall under a generic `gitError` type so we cannot tell the difference between a `git remote already exists` error and a `not a git repository` error.

Joshua-Anderson: In this case it isn't possible without larger rewrite. We could map STDERR parsing to errors in `git` package in the future, but I would like to rethink the git package in the future.

arschles: ok, is there an issue for the rethink-git-package work?

Joshua-Anderson: #196 is an error that is coming up due to differences in the git versions, but I'm kicking around the idea of using a full git go package to replace our homegrown client, but it's still in the feasibility stage right now.

PR #226, opened 2016-09-08
fix(whitelist): format the deis whitelist properly

kmala:


Joshua-Anderson:
Is this still  targeted for the 2.5 release?

kmala:
i would want it if possible but will check with @mboersma and @bacongobbler as they are doing release and fine with moving to next release as it is just an UX issue.

ultimateboy:
Before
```bash
$ deis whitelist:list
=== taller-addendum Whitelisted Addresses1.2.3.4
```
after
```bash
$ deis whitelist:list
=== taller-addendum Whitelisted Addresses
1.2.3.4
```

bacongobbler:
let's cut a v2.5.1.

PR #230, opened 2016-09-08
Limit tests

ultimateboy:
Fixes deis/workflow-e2e#196
Fixes deis/workflow-e2e#198
Fixes deis/workflow-e2e#199
Fixes deis/workflow-e2e#200
Fixes deis/workflow-e2e#201

### DIFF
--- a/cmd/users_test.go
+++ b/cmd/users_test.go
@@ -67,3 +67,46 @@ test
 *jkirk
 `, "output")
 }
+
+func TestUsersListLimit(t *testing.T) {
+	t.Parallel()
+	cf, server, err := testutil.NewTestServerAndClient()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer server.Close()
+	var b bytes.Buffer
+	cmdr := DeisCmd{WOut: &b, ConfigFile: cf}
+
+	server.Mux.HandleFunc("/v2/users/", func(w http.ResponseWriter, r *http.Request) {
+		testutil.SetHeaders(w)
+		fmt.Fprintf(w, `{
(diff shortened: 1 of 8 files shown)


cmd/users_test.go
```diff
@@ -67,3 +67,46 @@ test
...
+					"is_active": true,
+					"date_joined": "2014-10-19T22:01:00.601Z",
+					"groups": [],
+					"user_permissions": []
+				}
+			]
+		}`)
+	})
+
+	err = cmdr.UsersList(-1)
```
Joshua-Anderson: Do you want `1` here instead of `-1`?

Joshua-Anderson:
One minor thing I noticed, otherwise LGTM.

ultimateboy:
Nice catch @Joshua-Anderson. I'm going to spend a bit of time investigating why that was passing as it was written. But fixed.

Joshua-Anderson:
I'm thinking it was passing because the SDK returned `1` item but the `count` on the object was set to two. The code doesn't look at the requested limit, it only looks at the effective limits (returned items vs total count).

Joshua-Anderson:
Could you possibly add limits tests for `KeysList()`?

Joshua-Anderson:
I also noticed that we still need tests for  limits on `app.Logs()`. If you want to want to wait and do these in a different PR that's perfectly fine as well.

ultimateboy:
Thanks @Joshua-Anderson. Somehow missed the keys test. Added.

Joshua-Anderson:
After my changes to codecov, you need to rebase. Sorry!

PR #232, opened 2016-09-09
tests(settings): adding tests for FindHome() and SetHome()

ultimateboy:


settings/home_unix_test.go
```diff
@@ -0,0 +1,36 @@
...
+func TestFindHome(t *testing.T) {
+	previousHome := os.Getenv("HOME")
+
+	expectedHomeDir := "/d/e/f"
+	os.Setenv("HOME", expectedHomeDir)
+
+	assert.Equal(t, FindHome(), expectedHomeDir, "output")
+
+	// Reset HOME.
+	os.Setenv("HOME", previousHome)
```
Joshua-Anderson: Is there any reason to reset at the end? ENV variables set in a program cannot effect your environment.

ultimateboy: I was getting other errors from other tests when these were not set back.

Joshua-Anderson: What tests are erroring? We should make sure that every test that depends on home bing set sets it themselves

ultimateboy:
```?   	github.com/deis/workflow-cli/pkg/webbrowser	[no test files]
--- FAIL: TestLoadSave (0.00s)
	settings_test.go:95: mkdir /a: permission denied
FAIL```

ultimateboy: Addressed the reason why tests were failing if home dir env vars were not reset back and removed that code.

ultimateboy:
This branch got out of hand with some rebasing. Closing in favor of https://github.com/deis/workflow-cli/pull/245

PR #241, opened 2016-09-13
chore(Dockerfile): update go-dev and go toolchain

### LINKED ISSUE #240: Error with binary compiled with go-1.7.0: "dial tcp: no suitable address found"
Any kind of operation with `workflow-cli-v2.5.1` is raising the following error:

```
$ deis apps
Error: Get http://deis.lundgreen.rkp.local/v2/apps/?limit=100: dial tcp: no suitable address found
$ deis version
v2.5.1
```

This seems to be introduced with new go version 1.7.0 and [fixed](https://github.com/golang/go/issues/16739) on go 1.7.1. 

By compiling with go-1.7.1 the NAT error doesn't happen anymore.

Full history of the troubleshoot can be found on [slack channel](https://deis.slack.com/archives/community/p1473785265001407).

mboersma:
See https://github.com/golang/go/issues?q=milestone%3AGo1.7.1
and https://github.com/deis/docker-go-dev/releases/tag/0.17.1

Closes #240.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/deis/go-dev:0.17.0
+FROM quay.io/deis/go-dev:0.19.0
 # This Dockerfile is used to bundle the source and all dependencies into an image for testing.
 
 RUN echo "deb http://packages.cloud.google.com/apt cloud-sdk-jessie main" \
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ COPY glide.lock /go/src/github.com/deis/workflow-cli/
 
 WORKDIR /go/src/github.com/deis/workflow-cli
 
-RUN glide install --strip-vcs --strip-vendor
+RUN glide install --strip-vendor
 
 COPY ./_scripts /usr/local/bin
 


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/deis/go-dev:0.17.0
+FROM quay.io/deis/go-dev:0.17.1
 # This Dockerfile is used to bundle the source and all dependencies into an image for testing.
 
 RUN echo "deb http://packages.cloud.google.com/apt cloud-sdk-jessie main" \
```
Joshua-Anderson: The new go-dev is based on xenail so this should be changed from `jessie`

Dockerfile
```diff
@@ -20,7 +20,7 @@ COPY glide.lock /go/src/github.com/deis/workflow-cli/
 
 WORKDIR /go/src/github.com/deis/workflow-cli
 
-RUN glide install --strip-vcs --strip-vendor
```
Joshua-Anderson: Any particular reason you're removing strip vcs? This has some nice space savings since this uses `deis/pkg` which depends on kuberenetes.

mboersma: `--strip-vcs` is the default in `glide` 0.12+ and it emits a deprecation warning now if you include it.

Joshua-Anderson: TIL! Awesome.

PR #243, opened 2016-09-14
tests(parser): adding tests for parser package.

ultimateboy:


ultimateboy:
Before
```ok  	github.com/deis/workflow-cli/parser	1.024s	coverage: 1.5% of statements```

After
```ok  	github.com/deis/workflow-cli/parser	1.111s	coverage: 70.4% of statements```

**Comment by ultimateboy** [comment]
Now: ```ok  	github.com/deis/workflow-cli/parser	1.181s	coverage: 80.4% of statements```

**Comment by kmala** [comment]
Unit tests for parser was supposed to also test the help commands https://github.com/deis/workflow-e2e/pull/308#r75555869. Do you want to tackle it separately or in this PR.

File: parser/parser_test.go [hunk 1]
```diff
@@ -0,0 +1,125 @@
+package parser
+
+import (
+	"io"
+	"time"
+
+	"github.com/deis/controller-sdk-go/api"
+)
+
+type FakeCommander interface {
```
ultimateboy: Totally right. Didn't need to, removed.

parser/parser_test.go
```diff
@@ -0,0 +1,125 @@
...
+	Print(...interface{}) (int, error)
+	Printf(string, ...interface{}) (int, error)
+	PrintErrln(...interface{}) (int, error)
+	PrintErr(...interface{}) (int, error)
+	PrintErrf(string, ...interface{}) (int, error)
+	Version(bool) error
+}
+
+// DeisCmd is an implementation of Commander.
+type FakeDeisCmd struct {
```
ultimateboy: I pushed a commit which better asserts the right arguments are calling the right functions by having the functions return an error of their expected arg.

Joshua-Anderson: That's a good idea and something I didn't think of! I was thinking about testing the arguments that the commander interface function is called with.

ultimateboy:
@kmala Thanks for the issue reference. I'll see what I can do about getting help text covered tomorrow.

ultimateboy:
@kmala In an attempt to keep this diff from growing too large. I'm going to save the testing of the help/usage text for another PR.

kmala:
@ultimateboy sure...do we have an issue for that so that it can be tracked.Anyway this PR looks good.

PR #250, opened 2016-09-19
fix(config): support CRLF config files

ultimateboy:


ultimateboy:
Manually tested using both LF and CRLF line endings in a `.env` file. Both worked as expected.

PR #251, opened 2016-09-26
fix(perms): user-only settings file perms

jgmize:
Settings files contain auth tokens and should not be world-readable.

### DIFF
--- a/settings/settings.go
+++ b/settings/settings.go
@@ -90,13 +90,13 @@ func (s *Settings) Save(cf string) (string, error) {
 		return "", err
 	}
 
-	if err = os.MkdirAll(filepath.Join(FindHome(), "/.deis/"), 0775); err != nil {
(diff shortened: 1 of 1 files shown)


aboyett:
Nice catch. LGTM

jgmize:
Note that this PR only addresses the creation of new files and directories. As a followup I recommend adding an automated permissions fix for existing files & directories, or failing that, release notes instructing users of previous versions to fix the permissions manually (`chmod 700 ~/.deis && chmod 600 ~/.deis/*.json` or equivalent).

Joshua-Anderson:
A note to that point could pretty easily be added to release notes.

Joshua-Anderson:
If anybody has the time, a second review on this would be nice. This is a moderately large problem and should be fixed ASAP

Issue #246, opened 2016-09-16
DNS resolution issues on Mac OS X when connected to a VPN


lachie83:
golang binaries which are cross-compiled (like deis) tend to have CGO disabled which means that they don’t use OS X native resolver. This is causing us some grief as it means that golang binaries aren’t able to resolve DNS via our VPN. See https://github.com/hashicorp/terraform/pull/5925 as one example.

Tested and confirmed that this fixes the issue on Darwin based systems.
```
CGO_ENABLED=1
go build -a -ldflags '-X github.com/deis/workflow-cli/version.Version=v2.5.1' -o ./_dist/deis .
```

Is there anyway we can get CGO_ENABLED set of binaries published with the release?

helgi:
Does it have to be compiled on an OS X machine? I don't think we have an OS X node in our CI setup, right now that is.

Joshua-Anderson:
Yeah, this probably requires an OS X box. In theory CGO cross compilation is possible, but there aren't any `darwin` compilers for `linux` that I know of.

mboersma:
In the long run, we could provision a Jenkins node from https://www.macstadium.com/ or http://www.macincloud.com/, or dedicate an office MacBook to this task. For the time being, we can build `CGO_ENABLED` binaries locally and publish them over those created by the Jenkins job through cross-compiling.

mboersma:
I attached a MacBook Pro with OS X Yosemite on it to Jenkins as node5-boulder. We can investigate changing the workflow-cli jobs to prefer building the darwin client there with `CGO_ENABLED`.

gerred:
Could something like https://github.com/karalabe/xgo help? I haven't tried it, but this will apparently cross-compile with cgo enabled. I'm not sure if that pulls us down into some rabbit hole of having to include Darwin headers in the compilation environment or if xgo sorts that out directly.

gerred:
Hot diggity, it certainly does pull down the MacOS SDKs!

https://github.com/karalabe/xgo/blob/master/docker/base/Dockerfile#L43

We could *technically* compile the deis CLI to iOS even with xgo. I'm betting this would work just fine.

mboersma:
Reworking the CI job is a bit finicky, so we're going to retarget automating this for the next Workflow release.

PR #252, opened 2016-09-27
feat(Makefile): allow CGO_ENABLED override

vdice:
Enables fix for https://github.com/deis/workflow-cli/issues/246


vdice:
This PR is no longer needed; an alternate solution was developed in https://github.com/deis/jenkins-jobs/pull/246; closing.

PR #257, opened 2016-10-03
feat(keys): add an optional name support for keys:add

### LINKED ISSUE #256: deis key:add needs to take in an optional name
Based on the work in https://github.com/deis/controller/pull/1098 now SSH keys are unique.

To deal with that the `deis key:add` needs to either take in an optional name or we need to make it smarter and have it auto append a `-2` type suffix

helgi:
This arg can be used interchangably with the key arg to keep backwards compatibility

That allows a user to do `deis keys:add ~/.ssh/secret.pub` and the name is detected or `keys:add magic ~/.ssh/magic.pub` or `deis keys:add travis` and then get the choice list to pick from

Closes #256

### DIFF
--- a/cmd/keys.go
+++ b/cmd/keys.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -63,7 +64,7 @@ func (d *DeisCmd) KeyRemove(keyID string) error {
 }
 
 // KeyAdd adds keys.
-func (d *DeisCmd) KeyAdd(keyLocation string) error {
+func (d *DeisCmd) KeyAdd(name string, keyLocation string) error {
 	s, err := settings.Load(d.ConfigFile)
 
 	if err != nil {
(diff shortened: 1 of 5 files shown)


parser/keys.go
```diff
@@ -69,9 +69,13 @@ func keyAdd(argv []string, cmdr cmd.Commander) error {
 	usage := `
 Adds SSH keys for the logged in user.
 
-Usage: deis keys:add [<key>]
+Usage: deis keys:add [<name>] [<key>]
+
+<name> and <key> can be used in either order and are both optional
```
bacongobbler: Sorry, jumped the gun on that review. It looks like we'll need to manually test this new CLI functionality to ensure backwards compatibility and if this statement is true.

So if someone does `deis keys:add ~/.ssh/id_rsa`, it should still work according to this statement, oui?

cmd/keys.go
```diff
@@ -72,6 +73,16 @@ func (d *DeisCmd) KeyAdd(keyLocation string) error {
 
 	var key api.KeyCreateRequest
 
+	// check if name is the key
+	if name != "" && keyLocation == "" {
+		// detect of name is a file
```
bacongobbler: I guess answers my own question. :+1:

PR #260, opened 2016-10-07
fix(ssh): support ed25519 SSH keys

seanknox:
Updated the SSH parser regex to allow for [ed25519 SSH keys](http://ed25519.cr.yp.to/). Also added (very simple) test for key that would fail the regex. 

I develop in other languages but this is the very first golang code I've ever written, so very open to any nits you may have.

### DIFF
--- a/pkg/ssh/ssh.go
+++ b/pkg/ssh/ssh.go
@@ -6,7 +6,7 @@ import (
 )
 
 var (
-	pubKeyRegex = regexp.MustCompile("^(ssh-...|ecdsa-[^ ]+) ([^ ]+) ?(.*)")
+	pubKeyRegex = regexp.MustCompile("^(ssh-rsa|ssh-dss|ssh-ed25519|ecdsa-[^ ]+) ([^ ]+) ?(.*)")
 )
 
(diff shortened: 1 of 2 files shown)


Joshua-Anderson:
Do the controller and builder support ed25519 keys?

seanknox:
@Joshua-Anderson AFAICT they do; I added additional controller testing in https://github.com/deis/controller/pull/1109 to help answer that question in the future. builder uses [crypto/ssh](https://godoc.org/golang.org/x/crypto/ssh#PublicKey) which supports Ed25519.

Joshua-Anderson:
Cool! It looks like this has some minor linter problems:

```
Manadatory Linters: These must pass
pkg/ssh/ssh_test.go:1::warning: file is not gofmted with -s (gofmt)
Optional Linters: These should pass
pkg/ssh/ssh_test.go:24:9:warning: range var keyAndId should be keyAndID (golint)
pkg/ssh/ssh_test.go:43:9:warning: range var keyAndId should be keyAndID (golint)
```

You can run `make test-style` to run them locally

seanknox:
Fixed—thanks for the heads up.

PR #263, opened 2016-10-23
feat(limits-cmd): accept new limits:set value type

zinuzoid:
Accept new cli format limits:set web=500m/2000m
Value is map to requests/limits parameter in Kubernetes resource management.
See https://github.com/deis/workflow/issues/562
e2e https://github.com/deis/workflow-e2e/pull/335
Requires https://github.com/deis/controller/pull/1118

### DIFF
--- a/parser/limits.go
+++ b/parser/limits.go
@@ -61,21 +61,24 @@ Options:
 
 func limitSet(argv []string, cmdr cmd.Commander) error {
 	usage := `
-Sets resource limits for an application.
+Sets resource requests and limits for an application.
 
 A resource limit is a finite resource within a pod which we can apply
-restrictions through Kubernetes. This limit is applied to each individual
-pod, so setting a memory limit of 1G for an application means that each
-pod gets 1G of memory.
+restrictions through Kubernetes. A resource request is used by Kubernetes scheduler
+to select a node that can guarantee requested resource. If provided only one value,
+it'll be default by Kubernetes as both request and limit. These request and limit
+are applied to each individual pod, so setting a memory limit of 1G for an application
+means that each pod gets 1G of memory. Value needs to be within 0 <= request <= limit
 
-Usage: deis limits:set [options] <type>=<limit>...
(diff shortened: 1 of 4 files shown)


Joshua-Anderson:
Overall looks good! Looks like `limits_tests.go` needs to be formatted with `gofmt -s` though. Thanks for the contribution!

zinuzoid:
Thanks, I just notice now that `make test` didn't include `test-style`. Not sure if there's a reason for that.
Anyway this one is green now.

zinuzoid:
@Joshua-Anderson Thanks, I just notice now that make test didn't include test-style. Not sure if there's a reason for that.
Anyway this one is green now.

parser/limits.go
```diff
@@ -141,7 +143,7 @@ Options:
 	}
 
 	app := safeGetValue(args, "--app")
-	limits := args["<type>"].([]string)
+	limits := args["<type>=<value>"].([]string)
```
kmala: Can this command be unchanged and unset both requests and limit as now this command looks similar to set and if user want's to unset an individual request he can do it using the set?

zinuzoid: yep, make sense. I'll revert.

zinuzoid: now I'm not sure why I made this change. Doesn't even make sense to myself :-/

kmala:
Tested manually and this looks good,

 deis limits:set web=10M/100M
Applying limits... done

=== syrupy-yeomanry Limits

--- Memory
web     10M/100M

--- CPU
Unlimited

deis limits:set web=100M
Applying limits... done

=== syrupy-yeomanry Limits

--- Memory
web     100M

--- CPU
Unlimited

kmala:
It will be nice to add details of setting the value in the new way https://github.com/deis/workflow-cli/blob/master/parser/limits.go#L78-L89

kmala:
Jenkins, Add to whitelist.

Issue #264, opened 2016-10-24
Change --ssl-verify default on register and login to true


jschneiderhan:
The current default is `false`, which could leave clients vulnerable to MitM attacks.

PR incoming shortly :)

bacongobbler:
@jschneiderhan this is something we cannot change by default as the default use case are users who have not installed SSL certificates on their cluster. Enabling it by default will create errors in those cases, breaking the dev approach and backwards compatibility.

bacongobbler:
If we had a secure-by-default cluster then sure, but that'd be a breaking change and would merit a v3 release.

jschneiderhan:
@bacongobbler ok - thanks for letting me know. Would you like me to close this issue or leave it open for consideration in a future release?

bacongobbler:
I'll bring up a more generic topic on making Workflow secure by default. Thanks for bringing this to our attention!

Let's close this issue once that issue exists. :)

Joshua-Anderson:
@bacongobbler How would this break clusters for users that don't have ssl certificates? Workflow defaults to http, so a dev cluster wouldn't hit this, would they?

bacongobbler:
Ah I thought it'd come into play with http so no, a dev cluster would not hit this.

@jschneiderhan my mistake. PR away!

jschneiderhan:
Looks like I forgot to close this when #265 was merged

PR #265, opened 2016-10-24
feat(auth): change --ssl-verify default to true

jschneiderhan:
Changes the `--ssl-verify` default from `false` to `true` on both the `auth:login` and `auth:register` actions. This change should be backwards compatible with clusters using HTTP (since no SSL verification is performed) and clusters with valid SSL setups. Previously created profiles for clusters with invalid or self-signed certificates should continue to work, but new ones will require an explicit `--ssl-verify false` flag to be passed on register and login.

See discussion on #264

### DIFF
--- a/parser/auth.go
+++ b/parser/auth.go
@@ -64,8 +64,8 @@ Options:
     provide a password for the new account.
   --email=<email>
     provide an email address.
-  --ssl-verify=false
-    disables SSL certificate verification for API requests
+  --ssl-verify=true
+    enables/disables SSL certificate verification for API requests
 `
 
 	args, err := docopt.Parse(usage, argv, true, "", false, true)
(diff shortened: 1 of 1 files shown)


parser/auth.go
```diff
@@ -78,9 +78,9 @@ Options:
 	username := safeGetValue(args, "--username")
 	password := safeGetValue(args, "--password")
 	email := safeGetValue(args, "--email")
-	sslVerify := false
+	sslVerify := true
 
-	if args["--ssl-verify"] != nil && args["--ssl-verify"].(string) == "true" {
+	if args["--ssl-verify"] != nil && args["--ssl-verify"].(string) == "false" {
 		sslVerify = true
```
Joshua-Anderson: I think you may have forgotten to update this line.

jschneiderhan: Oops. Sorry about that. Thank you!

parser/auth.go
```diff
@@ -103,7 +103,7 @@ Options:
   --password=<password>
     provide a password for the account.
   --ssl-verify=false
```
ultimateboy: This line also needs updated.

jschneiderhan: Damn. Sorry. Updated.

bacongobbler:
let's slate this in for v2.9 so we can test this more in-depth.

jschneiderhan:
👍  Sure. I'm in no rush! We're just explicitly setting the flag to true in the meantime.

PR #269, opened 2016-11-07
feat(parser): add --login=true to auth:register

### LINKED ISSUE #266: Allow user registration without automatic login
If a cluster is run in admin_only registration mode, it would be great if `deis register` had a `--skip-login` flag to allow creating users as an admin without logging into their account. Alternatively a `users:create` action would make sense, that does not require passing the controller url.

bacongobbler:
This flag allows users to enable/disable automatic login after registering a user.

closes #266

### DIFF
--- a/cmd/auth.go
+++ b/cmd/auth.go
@@ -72,8 +72,7 @@ func (d *DeisCmd) Register(controller string, username string, password string, 
 
 	d.Printf("Registered %s\n", username)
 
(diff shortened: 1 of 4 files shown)


bacongobbler:
disregarding codecov because it's reporting on old code. Merging!

felixbuenemann:
Thanks!

PR #275, opened 2016-11-28
feat(cmd/labels.go): sort labels when listing

vdice:
also ref https://github.com/deis/controller-sdk-go/pull/107


vdice:
full ci/e2e run results will be from https://github.com/deis/workflow-e2e/pull/341

Issue #276, opened 2016-12-01
master branch not building; codecov issue?


vdice:
see https://ci.deis.io/job/Deis/job/workflow-cli/job/master/161/console

```
x> No CI provider detected.
    project root: .
--> token set from env
not found!
    Yaml found at: ./codecov.yml
--> slug set from yaml
==> Running gcov in . (disable via -X gcov)
==> Searching for coverage reports in:
    + .
    -> Found 1 reports
==> Python coveragepy not found
==> Detecting git/mercurial file structure
==> Reading reports
    + ./coverage.txt bytes=94388
==> Appending adjustments
    http://docs.codecov.io/docs/fixing-reports
    + Found 4707 adjustments
==> Uploading reports
    url: https://codecov.io
    query: branch=&commit=c9fb8303f35c37567a66df87bf905368760b4dd0&build=&build_url=&tag=&slug=deis%2Fworkflow-cli&yaml=.%2Fcodecov.yml&service=&flags=&pr=master&job=
    -> Pinging Codecov
HTTP 400
pr must match pattern ^(\d+|false|null|undefined|true)$
```

bacongobbler:
Definitely sounds like a codecov issue.

PR #277, opened 2016-12-01
fix(Jenkinsfile): set git_branch to env.BRANCH_NAME

vdice:
Our logic for discerning the [git branch](https://github.com/deis/workflow-cli/blob/master/Jenkinsfile#L55) was operating on the assumption that it would either be `remotes/origin/master` or a PR branch.  In fact, if the HEAD commit on the master branch is tagged, it will resolve to `tags/<tag>`.

As the logic above is used in many spots, this changes over to relying on `env.BRANCH_NAME`, which will be `master` even if tagged.  

Tested by porting this change manually to the master job to success: https://ci.deis.io/job/Deis/job/workflow-cli/job/master/173/ 

(With previous code, CI was executing logic on the assumption that it was a PR - kicking off pr test job, not building certain binaries, etc. - now we see it following the intended `master` path)

Fixes https://github.com/deis/workflow-cli/issues/276

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -52,11 +52,11 @@ stage 'Git Info'
 node(linux) {
 	checkout scm
 
-	git_branch = sh(returnStdout: true, script: 'git describe --all').trim()
+	git_branch = env.BRANCH_NAME
 	git_commit = sh(returnStdout: true, script: 'git rev-parse HEAD').trim()
 	git_tag = sh(returnStdout: true, script: 'git describe --abbrev=0 --tags').trim()
 
-	if (git_branch != "remotes/origin/master") {
+	if (git_branch != "master") {
 		// Determine actual PR commit, if necessary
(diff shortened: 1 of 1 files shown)


vdice:
Actually, realizing this isn't particular to codecov.  Anywhere we check the git branch in a binary `either 'remotes/origin/master' or <PR>` needs to be updated...

bacongobbler:
So with this change, what is the expected workflow when we make a PR to workflow-cli? Since `master` is hardcoded, will the PR job always fail until someone tells CI to build the PR as the expected branch?

If it's a matter of us uploading our reports to codecov as `remotes/origin/master`, would it be easier to strip `remotes/origin/` before uploading?

vdice:
@bacongobbler PRs will still work as they have been (you can see the job triggered by this PR [here](https://ci.deis.io/job/Deis/job/workflow-cli/view/Pull%20Requests/job/PR-277/9/console)).

This just changes to using the Jenkins Pipeline-provided `env.BRANCH_NAME` for resolving the branch name (it will be `master` for master and `PR-<number>` for a PR) instead of using the `git` cli to do so.

PR #278, opened 2016-12-02
fix(login): User should login by default when he registers

kmala:


parser/auth.go
```diff
@@ -81,24 +81,21 @@ Options:
...
-	// 2) calling `deis auth:register --login true`
-	if args["--login"] == nil || args["--login"].(string) == "true" {
-		return cmdr.Login(controller, username, password, sslVerify)
+	// 2) calling `deis auth:register --login false`
+	if args["--login"] != nil && args["--login"].(string) == "false" {
+		login = false
 	}
 
-	return nil
+	return cmdr.Register(controller, username, password, email, sslVerify, login)
```
bacongobbler: can we keep the same logic but refrain from calling `.Login` within `.Register`? for example:

```
login := true

if args["--login"] != nil && args["--login"].(string) == "false" {
    login = false
}

if err := cmdr.Register(controller, username, password, email, sslVerify); err != nil {
    return err
}

if login {
    return cmdr.Login(controller, username, password, sslVerify)
}

return nil
```

It prevents us from adding yet another feature flag to `cmdr.Register`, which we would need to add additional tests to account for that.

kmala: the username and password are captured in the cmdr.Register method, so we need to either pass them back or use string pointer for the above code to work which would change the Register signature anyway. I am ok with even doing that way.

PR #280, opened 2016-12-14
fix(healthcheck): Parse arguments properly for httpheaders

### LINKED ISSUE #279: Custom HTTP-Headers for Readiness-Checks ignored
Custom HTTP-Headers are ignored when trying to create a readiness check on Workflow v2.9.0 on K8s 1.4.6 with CLI v2.9.0-eb21ef2 (current master).

**Example:**
```sh
deis healthchecks:set readiness httpGet 5000 \
  --path /healthz \
  --type web \
  --initial-delay-timeout=20 \
  --timeout-seconds=5 \
  --header="X-Forwarded-Proto:https"
```

This is using the format listed in the specs, I also tried with `--header="X-Forwarded-Proto=https"`, which is the format listed in the docs.

The generated readiness probe in the app's deployment looks like:

```yaml
readinessProbe: 
  failureThreshold: 3
  httpGet: 
    path: /healthz
    port: 5000
    scheme: HTTP
  initialDelaySeconds: 20
  periodSeconds: 10
  successThreshold: 1
  timeoutSeconds: 5
```

Note that `httpHeaders` is missing.

The *expected* readiness probe would look like:

```yaml
readinessProbe:
  failureThreshold: 3
  httpGet:
    httpHeaders:
    - name: X-Forwarded-Proto
      value: https
    path: /healthz
    port: 5000
    scheme: HTTP
  initialDelaySeconds: 20
  periodSeconds: 10
  successThreshold: 1
  timeoutSeconds: 5
```

Prettified dump of the JSON I captured from the POST /v2/apps/:name/config/ request using tcpdump:

```json
{
  "healthcheck": {
    "web": {
      "readinessProbe": {
        "initialDelaySeconds": 20,
        "timeoutSeconds": 5,
        "periodSeconds": 10,
        "successThreshold": 1,
        "failureThreshold": 3,
        "httpGet": {
          "path": "/healthz",
          "port": 5000
        }
      }
    }
  }
}
```

There's no header info here either.

Btw. it would be really helpful if the correct format for specifying headers would be shown when running `deis healthchecks:set --help`.

kmala:
fixes #279 

requires https://github.com/deis/controller-sdk-go/pull/110

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,10 +1,10 @@
-hash: 1c11ae5822232e539faaf103162502905be42f0edf935dfb12012b83752a734f
-updated: 2016-11-29T10:19:52.552145656-07:00
+hash: 6ae1b56f1a4697e0cb1d61b22236897f7678dd2ae32a531948ca62ab68c5c6bb
+updated: 2016-12-14T12:52:24.333134278-07:00
 imports:
 - name: github.com/arschles/assert
   version: bb58b908265b5931732079df03f2818bf2167ba0
 - name: github.com/deis/controller-sdk-go
-  version: e4abe62d2accc3168b82ff54cb7e8aff2764d03b
+  version: 578082f82b3460b47d3c7e127313352f7117edfc
   subpackages:
   - api
   - apps
@@ -32,7 +32,7 @@ imports:
 - name: github.com/goware/urlx
   version: 8bb4a2e4339f55b15164907177e96e9faf885504
 - name: github.com/mattn/go-runewidth
-  version: d6bea18f789704b5f83375793155289da36a3c7f
+  version: 737072b4e32b7a5018b4a7125da8d12de90e8045
 - name: github.com/olekukonko/tablewriter
   version: bdcc175572fd7abece6c831e643891b9331bc9e7
 - name: github.com/PuerkitoBio/purell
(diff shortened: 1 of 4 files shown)


bacongobbler:
the `...` in the docstring means that you can supply this argument multiple times, so the usage should be `--header=X_DEIS_IS:awesome --header=X_MY_APP_IS_ALSO:awesome`. Does that not work?

bacongobbler:
From docopt's README:

> ... (ellipsis) one or more elements. To specify that arbitrary number of repeating elements could be accepted, use ellipsis (...), e.g. my_program.py FILE ... means one or more FILE-s are accepted. If you want to accept zero or more elements, use brackets, e.g.: my_program.py [FILE ...]. Ellipsis works as a unary operator on the expression to the left.

> If your usage patterns allows to match same-named option with argument or positional argument several times, the matched arguments will be collected into a list:

> ```
> Usage: my_program.py <file> <file> --path=<path>...
> ```

> I.e. invoked with `my_program.py file1 file2 --path=./here --path=./there` the returned dict will contain `args['<file>'] == ['file1', 'file2']` and `args['--path'] == ['./here', './there']`.

kmala:
no they aren't working for the options but are working for the arguments.

bacongobbler:
Any way we can fix this upstream in [docopt.go](https://github.com/docopt/docopt.go) instead and rev the version here? @mboersma is the core maintainer so I'm sure we could sneak him a cookie or something to motivate him ;)

mboersma:
Love to see a PR patch for docopt-go! :+1:

kmala:
i don't think we can have more than one arguments with ellipses(...) because the values are separated by space which means it can't differentiate mapping between the values and the args if there are more

bacongobbler:
Sounds good. This LGTM with the controller-sdk bump with the fix as well. 👍

Issue #271, opened 2016-11-16
use go-rootcerts to support custom PKI on darwin


tmc:
see https://github.com/hashicorp/go-rootcerts

bacongobbler:
Until go 1.8 is out this sounds like a good workaround. If you feel like making a PR, this would be a good opportunity.

bacongobbler:
this will be fixed in Go 1.7.4: https://groups.google.com/forum/#!topic/golang-announce/2lP5z9i9ySY

mboersma:
Fixed in #281.

PR #281, opened 2016-12-14
chore(Dockerfile): update go-dev to v0.21.0

mboersma:
Includes an OS X cert settings fix. Thanks to @bacongobbler for pointing this out.

See https://github.com/golang/go/issues?q=milestone%3AGo1.7.4


bacongobbler:
closes https://github.com/deis/workflow-cli/issues/271

PR #287, opened 2017-01-18
feat(Makefile): set docker build flags via environment variable

mboersma:


Joshua-Anderson:
I don't know the full context behind this, but it will have a significant impact on development time for `make test`... from seconds to several minutes.

mboersma:
@Joshua-Anderson it's true, which is why we've avoided doing this for so long, but building with cached Docker layers just means once in a while a bug is hidden and you don't discover it until later. In this case, switching the base image from `ubuntu-slim` to `ubuntu` broke a `gunzip` command for most images, but we didn't notice until building locally because a later Docker image layer was already cached in CI.

For developers, you can get the old behavior with:
```
DOCKER_BUILD_FLAGS= make docker-build
```

mboersma:
See this comment https://github.com/deis/postgres/pull/180#issuecomment-273562141 for more details.

vdice:
@mboersma when you get a chance, can you amend the commit and push again to kick of a fresh CI run?

mboersma:
Jenkins, test this please.

PR #288, opened 2017-01-19
feat(run-cmd): Allow using an unreleased image

robholland:
This is useful for build steps which must be completed before an app is fully
deployed to deis, for example asset compilation.

Builds on: https://github.com/deis/controller/pull/1208


robholland:
This PR fails CI as it depends on https://github.com/deis/controller-sdk-go/pull/115

PR #299, opened 2017-04-01
feat(cmd): add config:list --diff

### LINKED ISSUE #293: Add config:diff
Having a command that shows you a diff of given config vars (either from file or directly) with the config that comes back from `deis config` would help a lot. This would show you what actually changes when you do a `deis config:push`.

bacongobbler:
closes #293 

Usage:

```
$ deis config:pull
$ diff <(deis config:list --diff) .env
$ vim .env
$ diff <(deis config:list --diff) .env
3c3
< HELLO=world
---
> HELLO=deis
```

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -17,7 +17,7 @@ import (
 )
 
 // ConfigList lists an app's config.
-func (d *DeisCmd) ConfigList(appID string, oneLine bool) error {
+func (d *DeisCmd) ConfigList(appID string, format string) error {
 	s, appID, err := load(d.ConfigFile, appID)
 
 	if err != nil {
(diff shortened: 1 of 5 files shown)


cmd/config.go
```diff
@@ -31,16 +31,23 @@ func (d *DeisCmd) ConfigList(appID string, oneLine bool) error {
 
 	keys := sortKeys(config.Values)
 
-	if oneLine {
+	var configOutput *bytes.Buffer = new(bytes.Buffer)
```
kris-nova: Just curious.. why are we `new`ing here?

cmd/config.go
```diff
@@ -31,16 +31,23 @@ func (d *DeisCmd) ConfigList(appID string, oneLine bool) error {
...
+
+	switch format {
+	case "oneline":
 		for i, key := range keys {
 			sep := " "
 			if i == len(keys)-1 {
 				sep = "\n"
 			}
-			d.Printf("%s=%s%s", key, config.Values[key], sep)
+			fmt.Fprintf(configOutput, "%s=%s%s", key, config.Values[key], sep)
```
kris-nova: Do we need a `\n` here?

cmd/config.go
bacongobbler: Just initializing the buffer as per the docs. Is there some better way to initialize a bytes.buffer?

PR #306, opened 2017-05-25
check if config comes from pipe and stat size is > 0

### LINKED ISSUE #305: Get rid of stdin for `deis config:push`
At the implementation of `deis config:push` [here](https://github.com/deis/workflow-cli/blob/master/cmd/config.go#L238), it uses `stdin` to fetch the buffer, it was causing me problems, as in this [issue](https://github.com/concourse/concourse/issues/1192), running the command inside a script which gets its inputs from stdin, I had to pipe it with `< dev/null` to get the correct result or it would give me `Creating config... ...Error: Unknown Error (409): {"detail":"user changed nothing - release stopped"}`.

Can something be made to correct it?

pfeodrippe:
Fixes #305


Joshua-Anderson (changes requested):
Thanks a lot for the fix!
These changes didn't quite work for me though :cry:

test.sh:
```
deis config:push -p filename
```

```
$ ./test.sh < test
# read from pipe, not filename
```

However, you could switch the order of checks, checking if there is a filename defined before reading from Stdin, and that should solve the problem

Also, `deis config:pull` uses very similar code, so it would be great if that was patched up too.

pfeodrippe:
@Joshua-Anderson Ok, I'll test better this weekend o/.
`deis config:pull` didn't give me any trouble, but I'll check on it too

pfeodrippe:
@Joshua-Anderson You can see [at this gist](https://gist.github.com/pfeodrippe/656e4d099eac1c38a61ce3b958c5b23e) the `test` cli, it tests multiple cases for the `deis config:push` inputs.

You can [download](https://s3.amazonaws.com/deis-test-pr-306/deis.zip) the dev deis binary zipped (it will be updated if needed) I've built for testing.

Maybe we could put these tests as part of the build tests (of course, changing the scripts accordingly for automation).

pfeodrippe:
And how could I give a default to docopt-go at help and override it at code? Maybe it can say to us if the path came from the terminal or from the default value. You can see I'm using a hacked name for the path (.no-file)

parser/config.go
```diff
@@ -176,9 +176,11 @@ Usage: deis config:push [options]
 
 Options:
   -a --app=<app>
-    the uniquely identifiable name for the application.
+    The uniquely identifiable name for the application.
   -p <path>, --path=<path>
-    a path leading to an environment file [default: .env]
```
Joshua-Anderson: Sorry for giving you bad advice here about checking filepath first. I completely forgot about the `.env` default. It would be nice if we didn't have to go down this route, as it feels a little hacky. 

So much for the easy solution :smile:. I guess we do need to find the correct Sdin mode. If I have some time I'll help you experiment with this this week.

As for similar code in config:pull, [here](https://github.com/deis/workflow-cli/blob/master/cmd/config.go#L186) it is. It uses Stout, which means that redirecting a bash scripts output would likely run into similar issues to this.

Thanks so much for working on this and I apologize again for giving you bad advice :cry:!

pfeodrippe:
Ok, man, no worries :) I'll try to work on it at the weekend.
Let's check at the config:pull too o/

vdice:
@pfeodrippe is work on this PR still in progress? The next release (v2.16) is currently scheduled for this week and would be great to get this fix in!

pfeodrippe:
@vdice I'll have more time only after 2017-07-17, sorry =\, but I'll keep working at this PR, maybe it will enter at v2.17?

vdice:
@pfeodrippe No problems; sounds great! 👍

mboersma:
@pfeodrippe do you think this is something you can finish up in time for the September v2.18 release, or should we close this?

Cryptophobia:
Would you like to make this PR towards the fork of Deis - Hephy? Here is the link: https://github.com/teamhephy/workflow-cli

PR #308, opened 2017-06-13
fix(cmd) Fix proctype regex for setting limits

felixbuenemann:
This fixes the proctype regex to match [the one used by deis/controller](https://github.com/deis/controller/blob/master/rootfs/api/serializers.py#L21).

This fixes an issue reported by @tylux on slack where setting limits for a proctype called "worker1" failed with:

```
Error: worker1=100M/350M doesn’t fit format type=#unit or type=# or type=#/#
```

I've also added a few test cases that check for good and bad proctypes.

### DIFF
--- a/cmd/limits.go
+++ b/cmd/limits.go
@@ -145,7 +145,7 @@ func parseLimits(limits []string) (map[string]interface{}, error) {
 }
 
 func parseLimit(limit string) (string, string, error) {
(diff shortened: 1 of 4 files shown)


felixbuenemann:
I'll add some test cases, this was done quickly in the browser.

felixbuenemann:
OK, I've added a few relevang test cases in 1f5e242. Should be good to go from my side.

felixbuenemann:
> Could you possibly fix the regex used [here](https://github.com/deis/workflow-cli/blob/master/cmd/ps.go#L145) as well?

Fixed in f82cf8373b2742ac4a354220dd25a458c9bf160d.

PR #309, opened 2017-06-13
feat(cmd): Support Base64 encoded SSH_KEY

### LINKED ISSUE #274: deis config should handle base64 encoded SSH_KEY
Currently there's a special case for handling `deis config:set SSH_KEY=/path/to/key`.

The problem with this is, that it is broken when using for example:

```
DEIS_PROFILE=staging deis config:pull -a myapp
DEIS_PROFILE=production deis config:push -a myapp
Error: Could not parse SSH private key:
 LS0tLS1CRUdJT…
```

It would be great if `deis config:set SSH_KEY` could autodetect wether a path or an already base64 encoded key is given.

felixbuenemann:
This adds a check to see if the value of SSH_KEY is already a Base64
encoded value, which is the case if "deis config:push" is used after
using "deis config:pull" on an app that already has SSH_KEY set.

This fixes #274.

### DIFF
--- a/cmd/config_test.go
+++ b/cmd/config_test.go
@@ -154,7 +154,8 @@ func TestConfigSet(t *testing.T) {
 		if r.Method == "POST" {
 			testutil.AssertBody(t, api.Config{
 				Values: map[string]interface{}{
-					"TRUE": "false",
+					"TRUE":    "false",
+					"SSH_KEY": "LS0tLS1CRUdJTiBPUEVOU1NIIFBSSVZBVEUgS0VZLS0tLS0=",
 				},
 			}, r)
 		}
@@ -166,6 +167,7 @@ func TestConfigSet(t *testing.T) {
 			"TEST":  "testing",
 			"NCC":   "1701",
 			"TRUE":  "false",
+			"SSH_KEY": "LS0tLS1CRUdJTiBPUEVOU1NIIFBSSVZBVEUgS0VZLS0tLS0=",
 			"FLOAT": "12.34"
 	},
 	"memory": {},
@@ -181,16 +183,17 @@ func TestConfigSet(t *testing.T) {
 	var b bytes.Buffer
 	cmdr := DeisCmd{WOut: &b, ConfigFile: cf}
 
-	err = cmdr.ConfigSet("foo", []string{"TRUE=false"})
+	err = cmdr.ConfigSet("foo", []string{"TRUE=false", "SSH_KEY=-----BEGIN OPENSSH PRIVATE KEY-----"})
 	assert.NoErr(t, err)
 
 	assert.Equal(t, testutil.StripProgress(b.String()), `Creating config... done
 
 === foo Config
-FLOAT      12.34
-NCC        1701
-TEST       testing
-TRUE       false
+FLOAT        12.34
+NCC          1701
+SSH_KEY      LS0tLS1CRUdJTiBPUEVOU1NIIFBSSVZBVEUgS0VZLS0tLS0=
+TEST         testing
+TRUE         false
 `, "output")
 }
 
--- a/cmd/config_test.go
+++ b/cmd/config_test.go
@@ -3,7 +3,9 @@ package cmd
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
 	"net/http"
+	"os"
 	"testing"
 
 	"github.com/arschles/assert"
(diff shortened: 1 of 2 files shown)


felixbuenemann:
I tested the new behavior manually, but was unable to add automated tests, because the current location of the ssh key parsing code is poorly chosen and makes it hard to test.

In order to add sensible tests the code should probably be moved into a separate `parseSshKey` function which would be called from `parseConfig` instead of `ConfigSet`.

Joshua-Anderson:
We have some public ssh key parsing logic [in a package here](https://github.com/deis/workflow-cli/tree/master/pkg/ssh), maybe add the private ssh key logic parsing here too?

Also, I might swap the logic, so you check if the key is base64'd before trying to read a file and parse it, whih feels a little less hacky to me.

Thanks for contributing!

felixbuenemann:
@Joshua-Anderson I've thought about this and I'm wondering wether this is the right approach.

The weird thing is that `deis config` lists all environment variables in the base64-decoded form, except `SSH_KEY`. If it was properly decoding `SSH_KEY` this workaround would not be needed at all.

felixbuenemann:
OK, I've looked through the code and the reason that the SSH_KEY variable is base64 encoded is probably compatibility with heroku buildpack behavior. It's kinda strange, cause a private key is already in pem format which also uses base64 encoding…

felixbuenemann:
@Joshua-Anderson I've extracted the parsing code into a new `parseSshKey` function and added unit test for it as well as an integration style test in `ConfigSet` which ensures the code is called.

I've chosen to not put the new code into the ssh package, because it's implementation is too closely coupled to the specific behavior in ConfigSet to be re-used somewhere else, but I could move it if you disagree.

If you prefer I can squash to a single commit once the review is positive.

felixbuenemann:
Oh and I also changed the SSH Private Key matching regex so it also matches ed25519 or other ssh keys converted to the new openssh key format as part of a *refactor* commit. I could move that into a single commit if it's a problem.

cmd/config_test.go
```diff
@@ -149,16 +183,17 @@ func TestConfigSet(t *testing.T) {
 	var b bytes.Buffer
 	cmdr := DeisCmd{WOut: &b, ConfigFile: cf}
 
-	err = cmdr.ConfigSet("foo", []string{"TRUE=false"})
+	err = cmdr.ConfigSet("foo", []string{"TRUE=false", "SSH_KEY=-----BEGIN OPENSSH PRIVATE KEY-----"})
```
Joshua-Anderson: Not sure this is quite what you want, since you're checking if SSH_KEY is a base64 encoded key or a filepath.

felixbuenemann: This is the integration style test, it just checks that SSH_KEY is mangled and send to the server during config:set. It's only there to ensure the `parseSshKey` code which is covered by the unit tests is called at all, which flavor of encoding (plain/path/base64) we use here doesn't really matter.

felixbuenemann:
Fixed all errors from `make test-style`.

cmd/config_test.go
Joshua-Anderson: Yeah, it just a little confusing looking over, since you're using a private key header and base64 encoding, a combo that doesn't go together in normal usage.

felixbuenemann: Why doesn't it go together in normal usage?

The current docs say that you can set SSH_KEY to either a path or a private key.

The `parseSSHKey` doesn't *really* parse the key, it only checks if it *appears* to be a key by looking at the first line of the key, which is what I'm doing here.

In reality openssh wil happily use a key that just contains the base64 encoded key without any headers/footers, but that way we couldn't differentiate between a single or double base64 encoded key… There's also the golang crypto ssh module's `ParseRawPrivateKey` for actual parsing, but it only supports dss/rsa/ecdsa keys and would exclude ed25519 keys, which are perfectly fine to use here.

Maybe you could be more specific in what you would like me to change about this test.

felixbuenemann:
@Joshua-Anderson Stil waiting on review feedback from you.

Joshua-Anderson:
Sorry about that. I was confused by the fact you check for ssh keys as well as ssh keys base64 encoded by deis, but I understand where you're coming from now. I'd like to manually test it, but code looks LGTM!

mboersma:
I've tested this change locally and it works as advertised.

